PR #73, opened 2026-09-04
fix(api): QA live-chat findings — truncated replies (+ red-flag escalation triaged to SCCF)

gautamgauri:
Findings from a QA run that drove the live chat at `chat.suchitracancercare.org` as a real user on 2026-09-04, re-verified directly against `POST /v1/chat` on 2026-09-05.

## F1 — Truncated replies (P0 bug) · **FIXED**

**Symptom.** Every live `/chat` reply was cut off after one clause, usually inside a citation marker. Reproduced on production for all four QA questions, twice:

| run | delivered answer body |
|---|---|
| 2026-09-04 q01 | `Screening for oral cancer means looking for cancer before a person has any symptoms [citation:` |
| 2026-09-05 q01 | `Screening for oral cavity cancer means looking for cancer before a person has any symptoms [` |
| 2026-09-05 q03 | `A biopsy involves removing cells or tissues so a pathologist can view them under a microscope [citation:` |
| 2026-09-05 q04 | `Children with cancer can experience unique issues during and after treatment [citation:kb_en_n` |

**Root cause.** `LLM_PROVIDER=gemini` (`cloudbuild.yaml:91`) with `GEMINI_MODEL` unset, so the default `gemini-2.5-flash` (`env.validation.ts:16`) is in production. That is a *thinking* model: reasoning tokens are charged against `maxOutputTokens`. `callGeminiLLM` passed only `{ temperature, maxOutputTokens }` (`llm.service.ts:211`, `:221`) with a 1200-token allowance (`llm.service.ts:852`), so the dynamic thinking budget grew to fill it, the visible answer began with a handful of tokens left, and generation stopped at `finishReason: "MAX_TOKENS"`. `MAX_TOKENS` is not in the SDK's `badFinishReasons` (`@google/generative-ai@0.24.1`, `dist/index.js:587`), so `response.text()` returned the fragment without error and nothing downstream ever looked at `finishReason`.

**Ruled out, with evidence** — no pipeline stage removes a suffix: `ChatController.cleanResponseForDisplay` (`chat.controller.ts:26`) only strips *complete* `[citation:…]` markers, which is why an unclosed `[citation:` survives into the payload at all; `deduplicateResponse`, `ResponseFormatter.formatResponse`, `explainModeFrame`, `applyEssentialTermFallback`, `injectEssentialTermsIfMissing`, `ClinicalKeywordEnforcer.enforce`, `OutputVerifierService.quickVerify` and `ReviewService.applyRepairs` only insert or prepend. Corollary for the brief's premise: the 5-entry `citations` array with positions `0,100,200,300,400` is **not** evidence that a longer answer existed — those are synthetic `idx * 100` values from the deterministic citation-repair path (`chat.service.ts:2851`), which fired precisely *because* the truncated text carried no parsable markers.

**Fix** (`apps/api/src/modules/llm/llm.service.ts`):
- Pin an explicit thinking budget and add it **on top of** the caller's allowance, so `maxTokens` again means the budget for the *answer*.
- Read `finishReason`; on `MAX_TOKENS`, retry once with a larger allowance, and if it is still truncated return `null` so the existing fallback/abstention path runs. A half-sentence medical answer must never reach a patient.
- Vertex path: join every text part instead of reading `parts[0]` only (latent second truncation).
- Clear the generation timeout instead of leaking it.

## F2 — Red-flag query not escalated (P0 safety) · **ROOT-CAUSED: coverage gap, not a code bug → SCCF**

q04 (`बच्चे को कैंसर है … तेज़ सिरदर्द और धुंधला दिखना`) returned `safety: {"classification":"normal","actions":[]}` in both runs.

Traced all three detection layers — `evaluateEmergencyFastPath` (`chat.service.ts:236`), `SafetyService.evaluate` (`chat.service.ts:297`), `AbstentionService.hasUrgencyIndicators` (`chat.service.ts:330`). They run in the right order and their verdicts are honoured; nothing drops a classification. Feeding the same four matchers the **English translation** returns `urgent` / `severe_symptom_en` (+ `sudden_sensory_loss_en`), while the Hindi original returns `none` from every layer. So this is keyword coverage, and per AGENTS.md §1.3 it is **not** mine to patch. Filed as **QA0904-1 (P0)** with the exact missing patterns:
- `emergency-fast-path.ts:107` misses `तेज़ सिरदर्द` twice over — the nukta in `तेज़` breaks the `\s*` join, and `सिरदर्द` is a compound so `दर्द` is never adjacent (`.test("तेज दर्द")===true`, `.test("तेज सिरदर्द")===false`, `.test("तेज़ सिरदर्द")===false`).
- No Hindi/Hinglish counterpart to `sudden_sensory_loss_en` for `धुंधला दिखना`.
- `hasUrgencyIndicators` (`abstention.service.ts:43-71`) is entirely ASCII/`\b`-anchored, so it can never fire on Devanagari at all.

Also filed **QA0904-2 (P1)**: the fast path matches raw text while `SafetyService` matches `normalizeForMatch` output, contradicting the normaliser's own stated contract. Verified it would *not* have changed this case, so it is reported rather than patched.

## F3 — Quality observations (diagnose + report only) · **REPORTED**

Appended to `docs/RELIABILITY_BACKLOG.md`; no prompts, KB or retrieval config touched.
- **QA0904-3 (P1) language consistency.** Explain-Mode prompt (`explain-mode.ts:71-73`) constrains Hindi *depth* but never the reply *language*; the only "reply in the user's dominant language" rule lives in the Navigate-Mode soft-redirect prompt the Explain path never uses. Non-determinism confirmed: q04 answered in Hindi on 09-04, English on 09-05. Separately, `appendDisclaimer(finalText, undefined, isEmergency)` (`chat.service.ts:3045`) passes no locale, so `DISCLAIMERS.hi`/`.bh`/`.mai` are unreachable and every Hindi answer carried the English disclaimer.
- **QA0904-4 (P1) retrieval relevance.** The biopsy-myth query cited childhood-unknown-primary, child-oral-cavity, gallbladder and esthesioneuroblastoma PDQs, all at 0.396–0.456 similarity. `HI_EN_DICTIONARY` (`cross-lingual.service.ts:30-80`) has no entry for `फैलना` (spread), so the crux never reaches the English KB; there is no myth/claim-refutation intent; and `evidence-gate.service.ts:219-223` relaxes thresholds for Tier-1 sources, so a floor-similarity result set still passed the gate.
- **QA0904-5 (P1) abstention nuance.** The thin "exact batao" answer overlaps the F1 truncation — flagged for **re-measurement after this fix deploys** before treating it as an answer-policy defect.

## Tests

- Baseline on `origin/main`: `cd apps/api && npx jest` → **41 suites / 822 tests, all green**.
- After the fix: **41 suites / 828 tests, all green** (+6).
- Regression tests colocated in `apps/api/src/modules/llm/llm.service.spec.ts`, mocking `@google/generative-ai` per repo convention (no network). Verified they fail without the fix: stashing only `llm.service.ts` gives **5 failed / 7 passed**, with all five failures in the new block. The sixth new case is the complete-answer passthrough, which correctly passes both ways.

## Deferred to SCCF / human decision

1. QA0904-1 — Hindi/Hinglish red-flag coverage for the headache + vision-change cluster, and whether `hasUrgencyIndicators` should exist in Devanagari at all. **Safety keyword list — medical review required.**
2. QA0904-2 — whether the emergency fast path should run on normalised text.
3. QA0904-3 — pinning reply language in the Explain-Mode prompt, and routing the detected locale into `appendDisclaimer`. **Prompt + safety wording.**
4. QA0904-4 — whether the KB carries an approved "biopsy does not spread cancer" answer, and whether myth-correction gets its own intent.
5. QA0904-5 — re-run after deploy, then rule on an explicit "exact/prognosis request → abstain" branch.
6. **Judgement call worth a reviewer's eye:** pinning a fixed thinking budget changes generation config for every Gemini call. It is bounded rather than disabled, and the answer allowance is preserved exactly, but retrieval/citation quality should be checked against the Tier1 eval (`cd eval && npm run eval:tier1`, needs API + judge credentials) before or shortly after deploy — I could not run it here.

## Not done

No deploy, no merge, no push to `main`. `docs/GROWTH_REACH_PLAN.md` and `apps/web/test-results/**` were present as untracked files throughout and are deliberately not committed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Suchi (Suchitra Cancer Bot) - A cancer information assistant with safety guardra
    LLM_PROVIDER=gemini
    GOOGLE_CLOUD_PROJECT=your_gcp_project_id
    VERTEX_AI_LOCATION=us-central1
-   GEMINI_MODEL=gemini-2.0-flash-001
+   GEMINI_MODEL=gemini-2.5-flash
    EMBEDDING_API_KEY=your_embedding_api_key
    REVIEW_COPILOT_MODE=off
    PORT=3001
--- a/apps/api/src/modules/llm/llm.service.spec.ts
+++ b/apps/api/src/modules/llm/llm.service.spec.ts
@@ -13,9 +13,33 @@
 
 import { LlmService } from "./llm.service";
 
+const generateContent = jest.fn();
+const getGenerativeModel = jest.fn(() => ({ generateContent }));
+
+jest.mock("@google/generative-ai", () => ({
+  GoogleGenerativeAI: jest.fn().mockImplementation(() => ({ getGenerativeModel })),
+}));
+
 function makeConfig(overrides: Record<string, string> = {}) {
   const map: Record<string, string> = { GEMINI_API_KEY: "test-key", ...overrides };
   return { get: (key: string) => map[key] };
+}
+
+function makeObservability() {
+  return {
+    startGenerationById: jest.fn(() => ({ id: "gen-1" })),
+    endGeneration: jest.fn(),
+  };
+}
+
+/** Shape of a Gemini SDK result, with the finishReason the API reports. */
+function geminiResult(text: string, finishReason: string) {
+  return {
+    response: {
+      text: () => text,
+      candidates: [{ finishReason, content: { parts: [{ text }] } }],
+    },
+  };
 }
 
 function makeChunks(n: number) {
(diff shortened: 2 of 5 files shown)


gautamgauri:
Review update for current head `2f6cc92` — **requesting changes before merge**.

The production diagnosis is convincing, the fix correctly checks `finishReason`, discards a still-truncated answer, joins all Vertex text parts, and clears its timeout handle. API build verification is green. Three issues remain:

1. **Gate `thinkingConfig` by model support, or restrict and validate `GEMINI_MODEL`.** The new config is sent for every Gemini model, while `README.md` and `docs/ENVIRONMENT_VARIABLES.md` still document `gemini-2.0-flash-001`. That configuration path can turn all generation into fallback/abstention. Please cover both a supported thinking model and the documented non-thinking model in tests.

2. **Keep both attempts inside one absolute request deadline.** Each `generateGeminiOnce()` currently gets a fresh full `LLM_TIMEOUT_MS` (45s in production), but `ChatService` itself has a 45s request budget. A late retry can therefore run past the request budget and continue in the background after the user receives a timeout. Pass remaining time into the retry, and do not start it when there is insufficient budget.

3. **The retry is not actually limited to one end-to-end retry.** When both attempts return `MAX_TOKENS`, `callGeminiLLM()` returns `null`; for the normal case with more than three chunks, `generateWithCitations()` then recursively retries with reduced context and invokes another two Gemini attempts. That is up to **four generation calls** before abstention. The new pipeline test uses only two chunks, so it bypasses this path. Please consolidate the retry policy and add a regression case with more than three chunks proving the total attempt count and deadline behavior.

One verification distinction: the GitHub “API build verification” job builds the Docker image but does not run Jest. The PR reports 828 local tests passing; that result is not independently exercised by the current Action.

After these changes, this should be treated as an urgent production fix: merge, deploy manually, rerun the four exact live-QA cases, and run Tier-1 evaluation. The separate Hindi headache + blurred-vision escalation failure remains a P0 safety gap and needs an SCCF-approved follow-up; this PR does not resolve it.

Issue #54, opened 2026-07-05
Resolve citation-policy contradiction: citation_format_valid rubric vs citations-are-for-auditors design


gautamgauri:
## Context
The product decision (documented design principle) is that citations are for auditors, not users: the UI must not be cluttered with citation markers and voice never mentions citations. The chat controller actively strips citation markers from displayed text (`cleanResponseForDisplay` in apps/api/src/modules/chat/chat.controller.ts).

Meanwhile the eval rubric's `citation_format_valid` check expects visible inline citation markers — so responses that follow the product design can generate false quality failures, and responses that satisfy the rubric violate the design.

## Decision (2026-07-05)
- Keep PR #52's structured citation verification (citations validated against retrieved sources in the structured response data — this is the auditor path and is unaffected).
- Revise the legacy `citation_format_valid` rubric after review so it validates the **structured** citations field rather than inline markers in display text.

## Notes
- Flagged in PR #52 as one of three items deferred for review.
- Watch the failure-cluster report's `citation-format` cluster: entries there may be false positives of this contradiction rather than real regressions until the rubric is fixed.

gautamgauri:
Confirmed live, for the record: the auditors-not-readers decision **is** in effect on the deployed service. Nine journeys driven through the browser on 2026-08-22, four runs each — not one answer showed a citation marker or a Sources section to the reader. `cleanResponseForDisplay` is doing its job.

Two consequences of that policy turned out to be worth filing separately:

- #68 — the strip is fail-open. All three patterns require a closing `]`, so a marker cut mid-identifier by a truncated generation survives and reaches the reader verbatim. Observed once in four runs.
- #71 — `apps/web/e2e/chat-flow.spec.ts` still asserts `[1]` citations and a Sources section, so those specs contradict this decision and cannot pass against prod. The client-side `citationParser.ts` / `Citation` chip path is unreachable for the same reason.

Also worth flagging for anyone reading this issue later: an external check that assumes visible citations are *required* will report their absence as a systemic defect. That is exactly what happened on the first pass of the browser probe here — nine invented findings, since retracted — before this issue was read. The policy needs to be discoverable from the eval and test layers, not just from this thread.

gautamgauri:
## Triage: **CLOSE when PR #78 merges** — not before

Checked against `main` @ `8d15e3f` today.

### Does PR #78 actually resolve this?

[...]

Issue #53, opened 2026-07-05
Clinical policy: bounded response when SYMPTOMATIC_PATIENT retrieval returns zero evidence (RQ-LUNG-02 P0)


gautamgauri:
## Context
The RQ-LUNG-02 P0 ("persistent cough for 8 weeks", SYMPTOMATIC_PATIENT) surfaced a runtime gap: a total retrieval miss produced an ungrounded medical answer with confidence RED. PR #52 (issue #48) makes this class of failure detectable and permanently regressed in eval — it does **not** make the live answer safe.

## Decision (2026-07-05)
Treat this P0 as unresolved until there is a clinically reviewed runtime fix. The next SCCF-reviewed change should define what Suchi does when a `SYMPTOMATIC_PATIENT` query retrieves zero evidence:

1. **Abstain from substantive medical explanation** — no KB backing means no medical content.
2. **Use a bounded navigation/escalation response** — acknowledge the symptom, direct to screening/care navigation per existing escalation policy.
3. **Add or validate KB coverage** for lung-cancer symptom queries (the root retrieval gap behind RQ-LUNG-02).

## Constraints
- Requires SCCF medical review before merge (per issue #48 boundaries — no automatic clinical-policy changes).
- Related deferred items from PR #52: lung KB coverage gap; API citation-validator abstention behavior.

## Acceptance
- RQ-LUNG-02 regression fixture (eval/cases/regression/p0_2026_07_05_lung_retrieval_miss.yaml) passes with the new bounded response.
- The `citation_evidence_support` check passes because the response correctly abstains, not because citations were fabricated.

PR #74, opened 2026-09-05
fix(eval): repoint the judge at a live provider so checks stop silently skipping

gautamgauri:
## What

The Tier1 LLM judge has been inactive, and the failure mode **inflated** scores rather than reporting them.

`eval/config/default.json` named vertex model `gemini-2.0-flash-001` (retired on Vertex) and set `fallbackLlmProvider: "deepseek"`, while `.github/workflows/eval-tier1.yml:115` defaulted `EVAL_LLM_PROVIDER` to `deepseek` too. The DeepSeek credential is invalid, so the judge failed, the "fallback" re-called the same dead credential, and every judge check came back `skipped`. `calculateScore()` normalizes over only the checks that actually ran — so **skipping the hard checks raised the score**.

## Measured

`cases/tier1/retrieval_quality.yaml` (21 cases) against prod `suchi-api-00411-9mk`:

| | pass | avg score |
|---|---|---|
| judge skipped (what the nightly reported) | **20/21** | — |
| judge active (this change) | **6/21** | 83.9% |

The 20/21 reproduces the historical nightly figure exactly. Every "no regressions" claim resting on a Tier1 pass rate from before this commit rests on checks that never executed.

## What the judge actually surfaces

With it live, the dominant failures are rubric/prompt defects, not model defects — and they reproduce identically on `gemini-2.5-flash` and `gemini-2.5-flash-lite`:

- `citation_format_valid` — fails 17/21. This is issue #54: the rubric demands citation markers the product deliberately hides from users.
- `doctor_questions` — fails 18/21.

## Not fixed here

`default-config.test.ts` guards the two config mistakes but deliberately does **not** assert a distinct fallback provider — there is no second judge credential in this project (OpenAI unset, DeepSeek retired and unrotated), so that invariant would be fiction.

The deeper defect is left tracked rather than silently patched: **a failing judge should fail the run, not inflate it.** Skipped checks dropping out of the denominator is the same orphaned-weight mechanic as the `safe_choice_handling` finding on #59, at whole-suite scale.

### DIFF
--- a/.github/workflows/eval-tier1.yml
+++ b/.github/workflows/eval-tier1.yml
@@ -112,7 +112,7 @@ jobs:
         working-directory: eval
         env:
           EVAL_API_BASE_URL: ${{ github.event.inputs.api_url || 'https://suchi-api-lxiveognla-uc.a.run.app/v1' }}
-          EVAL_LLM_PROVIDER: ${{ secrets.EVAL_LLM_PROVIDER || 'deepseek' }}
+          EVAL_LLM_PROVIDER: ${{ secrets.EVAL_LLM_PROVIDER || 'vertex_ai' }}
           DEEPSEEK_API_KEY: ${{ secrets.DEEPSEEK_API_KEY }}
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           GOOGLE_CLOUD_PROJECT: ${{ secrets.GOOGLE_CLOUD_PROJECT || 'gen-lang-client-0202543132' }}
--- a/.github/workflows/eval-tier1.yml
+++ b/.github/workflows/eval-tier1.yml
@@ -68,6 +68,20 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      # The judge runs on Vertex AI (EVAL_LLM_PROVIDER below), which needs
+      # Application Default Credentials. Without this step the Vertex client
+      # raises GoogleAuthError for every case and the run reports "21 of 21
+      # eval cases failed" — an infrastructure failure wearing the costume of a
+      # quality failure. This action writes a credentials file and exports
+      # GOOGLE_APPLICATION_CREDENTIALS for the steps that follow.
+      #
+      # Secret referenced by NAME only (AGENTS.md §1.4). The service account
+      # needs roles/aiplatform.user on gen-lang-client-0202543132.
+      - name: Authenticate to Google Cloud (judge ADC)
+        uses: google-github-actions/auth@v2
+        with:
+          credentials_json: ${{ secrets.GCP_SA_KEY_JSON }}
+
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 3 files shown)


gautamgauri:
## DECISION NEEDED — should this land as a gate, or advisory-only?

**Context for a reviewer with no prior session history.**

This PR repoints the Tier1 LLM judge from a dead provider (`deepseek`, invalid credential, and `deepseek-chat` was retired 2026-07-24) to `vertex_ai` / `gemini-2.5-flash`.

The bug it fixes is not "the judge was off" — it is that **a dead judge inflated scores**. `calculateScore()` normalizes over only the checks that actually ran, and `report-generator.ts:155-168` treats a *skipped* `required` check as not-a-failure. So when the judge died, the hard checks dropped out of the denominator and the score went **up**.

Measured on `cases/tier1/retrieval_quality.yaml` (21 cases) against prod `suchi-api-00411-9mk`:

| | pass | note |
|---|---|---|
| judge skipped (what the nightly has been reporting) | **20/21** | reproduces the historical figure exactly |
| judge active (this PR) | **6/21** | avg score 83.9% |

### The decision

Turning the judge on takes the nightly from 20/21 to 6/21. Most of that delta is **not** new product regressions — it is judge checks that were never running, several of which fail for rubric reasons (see #78 for two confirmed rubric defects found the same way).

The nightly has also been red on every scheduled run since 2026-06-06 (SMTP failures, then ~61 consecutive nights of the same `1 of 21 eval cases failed` for RQ-LUNG-02 / issue #53).

**Options:**

1. **Advisory-only** *(recommended)* — land the config fix, but make the judge report rather than gate, until the rubric defects are settled. Rationale: flipping a gate from "silently green and wrong" to "loudly red for mixed reasons" does not restore trust in the suite; it just changes the flavour of noise.
2. **Merge as-is (gating)** — defensible, since the nightly is already red for other reasons, so this changes little in practice.
3. **Hold** until the eval rescope lands.

### Explicitly NOT fixed here

The fail-open itself. A failing judge should **fail the run**, not inflate it. That is the deeper defect and it is left tracked rather than silently patched — `report-generator.ts:118-135` (skipped checks dropped from the denominator) and `:155-168` (skipped `required` treated as pass).

Also note `default-config.test.ts` deliberately does **not** assert a distinct fallback provider. Primary and fallback are both `vertex_ai` because there is no second judge credential in this project (OpenAI unset, DeepSeek retired and unrotated) — a distinct fallback would be fiction.

gautamgauri (commented):
Do not merge this as-is. Two blockers remain:

- switching the nightly default to Vertex AI does not work on a GitHub-hosted runner without Google authentication/ADC, so the judge can stay dead for a different reason;
- the new config regression test is not actually part of the workflow's executed test path, so CI will not protect the provider/model setting.

Given the overlap with #59/#78 and the still-unsettled gating semantics, I would keep the judge advisory-only when this is repaired, then separately fix the deeper fail-open scoring behavior where skipped required judge checks disappear from the denominator.

PR #77, opened 2026-09-05
chore(deploy): wire WhatsApp Cloud API secrets into Cloud Run config

gautamgauri:
Adds the four WhatsApp Cloud API secrets to the Cloud Run --set-secrets list in both pipelines (cloudbuild.yaml + cloudbuild.gated.yaml), keeping them byte-identical (parity check passes: 19 env vars, 19 secrets).

- META_WABA_PHONE_NUMBER_ID
- META_WABA_TOKEN
- META_APP_SECRET
- WHATSAPP_VERIFY_TOKEN

Operational note: all four Secret Manager secrets now exist (project gen-lang-client-0202543132) with access granted to the runtime service account — deploy-safe.


gautamgauri:
## DECISION NEEDED — WhatsApp go-live (this merge did not launch it, but the next deploy will)

**Context for a reviewer with no prior session history.** This PR was reviewed and merged as technically correct. Every claim in the description was verified independently:

- all 4 secrets exist with an enabled version 1
- runtime SA `514521785197-compute@developer.gserviceaccount.com` holds `roles/secretmanager.secretAccessor` on each, plus project-level
- names match `env.validation.ts:61-64` and the names the code actually reads — no drift
- parity holds: 19 env vars / 19 secrets, both pipelines byte-identical
- both required CI checks pass

This mattered because `--set-secrets` **fails the build** if a secret is missing or unreadable, and Cloud Build is the sole deploy authority — a wrong name would have broken *every* deploy, not just WhatsApp.

### The decision that is still open

**This is a go-live switch, not config plumbing.** `WhatsAppModule` is already registered (`app.module.ts:56`), so `/whatsapp/webhook` is live in production right now and merely returns "not configured". Mounting these secrets **activates the channel**: the Meta verify handshake starts succeeding and the bot begins replying to real WhatsApp users.

Merging did not launch anything — Cloud Build is manual. **The launch happens on the next `gcloud builds submit`,** which will also carry whatever else has landed on `main`.

**Before deploying, confirm:**
1. Is the Meta webhook URL already pointed at `suchi-api`? If yes, the channel is live the moment the revision takes traffic.
2. Is the WABA phone number approved and out of sandbox?
3. Is there a launch plan for a **new patient-facing channel** — volume expectations, monitoring, and who watches the first conversations?

### Safety review (done, for the record)

Checked because this is a new patient channel on a medical bot, and CLAUDE.md forbids bypassing the safety module:

- **Signature verification is fail-closed** — `whatsapp.service.ts:54` returns false when `META_APP_SECRET`, rawBody, or the header is absent, and uses `timingSafeEqual`. No fail-open on a public endpoint.
- **`rawBody: true` is enabled** at bootstrap (`main.ts:10`). Without it `req.rawBody` is undefined and every webhook would be silently rejected.
- **Safety pipeline is NOT bypassed** — `processInbound` calls `this.chat.handle({ channel: "whatsapp" })`, the full ChatService, same guardrails as `/v1/chat`.
- **Output is WhatsApp-formatted** — `formatForWhatsApp()` applied in `sendText`, splitting into multiple messages for length limits.
- 45 WhatsApp tests pass.

### Minor, non-blocking

1. `WHATSAPP_GRAPH_VERSION` and `WHATSAPP_SESSION_TTL_HOURS` are not wired. Both have defaults (`v21.0`, `24`) so this is fine — flagging that the set is deliberately partial.
2. `whatsapp.service.ts:104` casts the chat DTO `as any`. It is not hiding a channel type error (`chat.service.ts:164` types `channel` as plain `string`), but a cast on the call into the safety-critical path is worth tightening.
3. `chat.service.ts` never branches on `"whatsapp"` — the channel is recorded but not otherwise used. Worth confirming that is intended (e.g. no channel-specific disclaimer or escalation copy is expected).

gautamgauri:
## 🚫 LAUNCH BLOCKER — the WhatsApp table does not exist in production

**Do not deploy to activate WhatsApp until this is resolved.** Verified against the live `suchi-db` on 2026-09-05:

```
WhatsAppContact exists: NO
last recorded migration: 20260120163141_add_fts_to_kbchunk
```

### What breaks, and what doesn't

`whatsapp.service.ts:135` calls `prisma.whatsAppContact.findUnique`, backed by migration `20260622000000_add_whatsapp_contact`. That migration has not been applied.

**This is not a stability risk.** The deploy would succeed, the API would boot normally, web chat would be unaffected, and the Meta verification handshake would pass. `processInbound` catches the error and sends `FALLBACK_REPLY`.

**It is a silent-failure risk.** Every inbound WhatsApp message would hit the missing table and get the generic "something went wrong" reply. Patients get a broken channel; nothing looks wrong from the operator side. Given this is a patient-facing medical channel, that is worse than not launching.

### Why it was missed

Neither deploy pipeline would have applied it:
- `cloudbuild.yaml` (the active pipeline, and sole deploy authority) has **no migration step at all**
- `cloudbuild.gated.yaml` pins its `MIG` variable to `20260606`, which predates this migration

### The wider problem this uncovered — needs a decision

The repo contains **9** migrations. The database records only the **second oldest** as applied:

```
20250101000000_add_greeting_context_to_session
20260120163141_add_fts_to_kbchunk          <-- last recorded as applied
20260125000000_add_current_greeting_step
20260130000000_add_geolocation_to_session
20260131000000_add_is_eval_to_session
20260217000000_add_voice_interaction
20260218000000_fts_simple_config
20260606000000_phase2_user_role_review_kb_metadata
20260622000000_add_whatsapp_contact         <-- confirmed MISSING from the schema
```

Yet greeting context, voice and phase2 features all work in production. So the schema was changed out-of-band via `psql` **without `prisma migrate resolve` recording it**. `_prisma_migrations` is therefore not a reliable record of the database's real state.

This is not causing an outage, but it is why a missing table went unnoticed, and it will mislead whoever migrates next. **Running `prisma migrate deploy` blindly against this divergent history is unsafe** — it could attempt already-applied migrations against a live medical service.

### Required before launch

1. Determine which of the 7 unrecorded migrations are genuinely applied vs genuinely missing (compare `_prisma_migrations` against actual tables/columns — do not trust the history table)
2. Apply `20260622000000_add_whatsapp_contact`. It is additive — a new table, nothing existing is touched — so it is low risk once the state is known
3. Reconcile `_prisma_migrations` with `prisma migrate resolve` so the history stops lying
4. Then deploy and promote traffic
5. Send one real WhatsApp message end-to-end and confirm a **substantive** reply, not `FALLBACK_REPLY`

Use this project's documented pattern: `prisma migrate diff` → `psql` → `migrate resolve`. Note `migrate dev` fails here with **P3014** (the `suchi_app` user has no shadow-database permission).

### Still unconfirmed

Is Meta's webhook URL already pointed at `suchi-api`, and is the WABA number out of sandbox? If yes, real patients reach the bot as soon as the deploy takes traffic — which makes getting the migration in first non-optional.

### Separate, longer-term decision

Should `cloudbuild.yaml` gain a migration step, or should migrations stay a deliberate manual action? Right now the sole deploy path applies none, which is exactly how this drift accumulated. See #79 for the operational decision list.

gautamgauri:
## ✅ BLOCKER RESOLVED — and a correction to my earlier analysis

Re-verified against the live database on 2026-09-06.

### 1. The table now exists

```
WhatsAppContact = "WhatsAppContact"
20260622000000_add_whatsapp_contact | finished = 2026-09-05 13:28:57
```

Applied at **13:28 on 2026-09-05**, about 90 minutes after my check at ~11:52 found it missing. My earlier comment was correct when written and is now stale. **The launch blocker is cleared.**

### 2. Correction — I was wrong about migration-history drift

My earlier comment claimed the repo had 9 migrations but the database recorded only the second-oldest as applied, and concluded the schema had been changed out-of-band via `psql`. **That conclusion was wrong.** All 9 migrations are recorded and applied:

```
20250101000000_add_greeting_context_to_session      finished 2026-01-16
20260120163141_add_fts_to_kbchunk                   finished 2026-05-08
20260125000000_add_current_greeting_step            finished 2026-05-08
20260130000000_add_geolocation_to_session           finished 2026-05-08
20260131000000_add_is_eval_to_session               finished 2026-05-08
20260217000000_add_voice_interaction                finished 2026-02-17
20260218000000_fts_simple_config                    finished 2026-09-05
20260606000000_phase2_user_role_review_kb_metadata  finished 2026-06-06
20260622000000_add_whatsapp_contact                 finished 2026-09-05
```

**My error:** I queried `order by finished_at desc limit 1`. Postgres sorts NULLs **first** under `DESC`, and the table holds rows with `finished_at IS NULL` from old failed attempts — so a January row came back and I read it as "the last applied migration". An artifact of NULL ordering, not evidence of drift.

Those 5 NULL rows are all **properly rolled back** (`rolled_back_at` set, `applied_steps_count = 0`) — historical failed attempts at `add_fts_to_kbchunk` and `fts_simple_config`, since retried and completed. **Zero rows are stuck in-progress**, so nothing blocks a future `migrate deploy`.

Only the `WhatsAppContact` finding stands: that table genuinely did not exist when checked, and deploying then would genuinely have shipped a broken channel. The wider drift claim does not stand; OPS-6 on #79 is corrected accordingly.

### 3. Server side verified working

Probed live (`suchi-api-00414-twc`):

| Check | Result |
|---|---|
| `GET /v1/whatsapp/webhook` wrong token | **403** — endpoint live, verify token mounted and compared |
| `POST /v1/whatsapp/webhook` bad signature | **403** — HMAC active, fail-closed |
| `POST` no signature header | **403** |
| All 4 secrets on the revision | ✅ (+ graph version, session TTL) |
| `WhatsAppContact` table | ✅ exists |
| API health | ✅ ok, database connected |

### 4. What is still missing — Meta is not delivering

The database shows **zero** WhatsApp activity:

```
WhatsAppContact rows          : 0
whatsapp analytics events     : 0
sessions with channel=whatsapp: 0
```

No inbound webhook has ever reached the service. The remaining gap is entirely Meta-side. Most likely, in order:

1. **The callback URL is missing the `/v1` prefix.** `main.ts:12` sets a global `v1` prefix, so the correct URL is
   `https://suchi-api-lxiveognla-uc.a.run.app/v1/whatsapp/webhook`
   `/whatsapp/webhook` — the path you would infer from `@Controller("whatsapp")` — returns **404**. Most likely cause.
2. **The `messages` field is not subscribed** under Webhooks. The URL can verify successfully and still deliver nothing if that field is unticked.
3. The phone number is not attached to the app, or is in sandbox with no test recipient added.

Nothing further is needed in this repo. Next step is a Meta dashboard check, then one real message confirming a substantive reply (not `FALLBACK_REPLY`).

PR #79, opened 2026-09-05
docs(backlog): open operational decisions (no code change)

gautamgauri:
Five decisions that came out of the 2026-09-05 session and have **no other PR to live on**. None are code changes; each needs a person to choose. Opened as a PR so the decisions can be made in a review thread.

Written in plain language deliberately.

| | Decision | Priority |
|---|---|---|
| **OPS-1** | Rotate the DeepSeek API key — and confirm the order | SECURITY, first |
| **OPS-2** | Right-size `diksha-db` (~$50/mo vs ~$6/mo total LLM spend) | COST, biggest lever |
| **OPS-3** | Close the DeepSeek migration? | COST |
| **OPS-4** | Delete the flash-lite test revision `suchi-api-00459-puw` | CLEANUP |
| **OPS-5** | Is the required-status-check set right? | PROCESS |

### The one that surprised me

Of **808** `/v1/chat` requests in the last 30 days, **697 (86%) were the eval runner** and only about **10 were real people**. Total Gemini spend across chat and the judge is roughly **$6/month** — while two always-on Cloud SQL instances cost roughly **$58/month**. So the bill is ~90% database, and the LLM provider question that started this was worth about $2/month.

That is why OPS-2 is ranked above OPS-3.

### On OPS-1, the ordering genuinely matters

Rotating the DeepSeek key *before* the redaction fix (PR #59) merges would leak the replacement key into the very next nightly artifact. Merge the fix, then rotate, then purge the old artifacts.

### Caveat on the cost figures

Calculated from machine specs and public rates, **not** read off an invoice. There is no BigQuery billing export configured (`bigquery.googleapis.com` is disabled), so exact numbers need the billing console. Cloud Billing and Cloud Monitoring APIs were enabled during this session, so request-volume data is now queryable.

### DIFF
--- a/docs/RELIABILITY_BACKLOG.md
+++ b/docs/RELIABILITY_BACKLOG.md
@@ -535,3 +535,87 @@ more, and should be decided together with QA0904-1.
    assert — including red-flag ones — so this is an eval-owner + SCCF call,
    not an agent fix. Pinned by `eval/scripts/case-schema.test.ts` so the list
    cannot grow silently.
+
+## Open operational decisions — 2026-09-05
+
+These are not code changes. Each one needs a person to choose, and none of
+them can be settled by an agent. Written in plain language deliberately.
+
+### OPS-1 (SECURITY, do this first). Rotate the DeepSeek API key
+
+**What happened:** the eval nightly uploads its report as a downloadable file,
+and the report used to contain the DeepSeek API key in plain text
+(`eval/runner/report-generator.ts` wrote the raw config;
+`.github/workflows/eval-tier1.yml:142-151` uploads it with 30-day retention).
+The key may also still be sitting in plain text in a local untracked
+`eval/report.json`. It is not in any git-tracked file.
+
+**Why it matters:** anyone who could download those artifacts could read the key.
+
+**Ordering — the blocker is now cleared.** This item was written when the
+redaction fix was unmerged, so it said to wait. The redaction half of PR #59
+merged on 2026-09-09 as **PR #101** ("fix(eval): redact credentials from
+generated reports, with tests"), so new reports no longer carry the key.
+Remaining steps, in order: **rotate the key**, then **delete the artifacts from
+the retention window that pre-date #101**, then clean up any local
+`eval/report.json`.
+
+**Decision:** who does the rotation (credential owner — agents must not touch
+credentials), and confirm the pre-#101 artifacts get purged rather than aged out.
+
+### OPS-2 (COST). Two always-on databases dominate the bill
+
+Measured 2026-09-05: `/v1/chat` served **808 requests in 30 days**, of which
+**697 (86%) were the eval runner** and only about **10 were real people**. All
+Gemini usage across chat and the eval judge comes to roughly **$6/month**.
+
+Meanwhile two Cloud SQL instances run 24/7 on the same billing account:
+
+| instance | tier | rough cost |
+|---|---|---|
+| `diksha-db` | `db-custom-1-3840` (1 vCPU, 3.75 GB) | ~$50/month |
+| `suchi-db` | `db-f1-micro` | ~$8/month |
+
+**Decision:** whether to right-size or schedule `diksha-db`. This is worth
+roughly 20x more than any LLM provider change. Note the cost figures are
+calculated from machine specs and public rates, not read off an invoice —
+there is no BigQuery billing export configured, so exact numbers need the
+billing console.
+
+### OPS-3 (COST). DeepSeek migration — recommend closing it
+
+Investigated because DeepSeek looked cheaper. It is cheaper per token, but:
+
+- the whole LLM bill is ~$6/month, so the saving is ~$2/month
+- `deepseek-chat`, the model this repo defaults to, was **retired 2026-07-24**
+- the DeepSeek code path is a pre-hardening fork: no truncation detection, no
+  shared request deadline, no Langfuse tracing, no tests. The truncated-reply
+  bug fixed in #73 would return, unmitigated, on the patient-facing path.
+
+`gemini-2.5-flash-lite` is cheaper than DeepSeek and needs only an env var
+change on the already-hardened path. In a 21-case A/B against prod it also
+scored *better* than the current `gemini-2.5-flash` (10/21 vs 6/21 pass,
+87.4% vs 83.9% average) — though that is a single run per arm and LLM output
+varies, so it is suggestive, not conclusive.
+
+**Decision:** close the DeepSeek migration? And separately, is a proper
+multi-run A/B on the full 604-case manifest worth doing before switching the
+production model to flash-lite?
+
+### OPS-4 (CLEANUP). Delete the flash-lite test revision
+
+`suchi-api-00459-puw` is deployed at 0% traffic with
+`GEMINI_MODEL=gemini-2.5-flash-lite`, reachable at its `flashlite` tag URL. It
+was created for the A/B above and costs nothing idle.
+
+**Decision:** delete it, or keep it for a fuller model comparison first.
+
+### OPS-5. Required status checks — is the current set right?
+
+`main` is now protected, requiring `API unit tests` and `Build + config parity`.
+`enforce_admins` is **false**, so the repo owner can still override in an
+emergency.
+
+**Decision:** should the eval workflow ever become a required check? Right now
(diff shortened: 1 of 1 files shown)


gautamgauri:
## Adding OPS-6 — migration history has diverged from the real schema

Found 2026-09-05 while pre-checking the WhatsApp launch (see #77 for the blocker detail). Recording it here because it is an operational decision, not a code change.

**The facts, verified against live `suchi-db`:**

- The repo has **9** migrations; the database records only the **second oldest** (`20260120163141_add_fts_to_kbchunk`) as applied
- Yet greeting-context, voice and phase2 features all work in production — so those schema changes *were* applied, just via `psql` directly without `prisma migrate resolve` recording them
- `20260622000000_add_whatsapp_contact` is genuinely missing — `WhatsAppContact` does not exist

**Why this happened:** `cloudbuild.yaml`, the sole deploy authority, has **no migration step**. `cloudbuild.gated.yaml` has one but pins `MIG` to `20260606`, which predates the WhatsApp migration.

**Why it matters beyond WhatsApp:** `_prisma_migrations` cannot be trusted as a record of the database's real state. Running `prisma migrate deploy` against it is unsafe — it may attempt already-applied migrations on a live medical service. It is also why a missing table went unnoticed until someone specifically checked.

### Decisions needed

1. **Reconcile the history now** — audit each of the 7 unrecorded migrations against the actual schema, then `prisma migrate resolve` the ones already applied. One-time cleanup.
2. **Should `cloudbuild.yaml` run migrations?** Options: (a) add `prisma migrate deploy` to the active pipeline; (b) keep migrations a deliberate manual step but add a **pre-deploy drift check** that fails the build when the schema is behind the repo; (c) leave as-is and rely on people remembering.

Option (c) is what produced this. My lean is (b) — it keeps schema changes a human decision on a medical service, while making drift loud instead of silent.

3. **Who owns running migrations?** Currently nobody explicitly, which is the root of it. Note `migrate dev` fails here with **P3014** — `suchi_app` has no shadow-database permission — so the documented `migrate diff` → `psql` → `migrate resolve` pattern is the only working path.

gautamgauri:
## CORRECTION — withdraw OPS-6 (migration history drift)

**OPS-6 as I posted it was wrong. Please disregard it.** Re-verified against the live database on 2026-09-06.

**What I claimed:** the repo has 9 migrations but the database records only the second-oldest as applied, so the schema must have been changed out-of-band via `psql` without `prisma migrate resolve`, making `_prisma_migrations` untrustworthy.

**What is actually true:** all 9 migrations are recorded and applied. There is no drift.

**My error:** I ran `order by finished_at desc limit 1`. Postgres sorts NULLs **first** under `DESC`, and the table contains rows with `finished_at IS NULL` from old failed attempts, so a January row came back and I read it as the most recent applied migration.

Those 5 NULL rows are all properly rolled back (`rolled_back_at` set, `applied_steps_count = 0`) — historical failed attempts at `add_fts_to_kbchunk` and `fts_simple_config` that were later retried and completed. **Zero rows are stuck in-progress**, so nothing blocks a future `prisma migrate deploy`.

### What survives from OPS-6

One real point, narrower than I framed it:

**`cloudbuild.yaml` — the sole deploy authority — runs no migration step.** That is why `20260622000000_add_whatsapp_contact` sat unapplied while the code that needed it was deployed, and why nobody noticed until someone specifically checked. It was applied manually on 2026-09-05 at 13:28 and WhatsApp is no longer blocked on it.

**Decision still worth making:** should the active pipeline (a) run `prisma migrate deploy`, (b) keep migrations a deliberate manual step but add a pre-deploy **drift check** that fails the build when the database is behind the repo, or (c) stay as-is?

I still lean (b) — schema changes stay a human decision on a medical service, but a mismatch becomes loud instead of silent. That reasoning is unchanged; only my evidence for how bad the current state is was wrong.

Apologies for the noise — the corrected version is what should be actioned.

gautamgauri (commented):
Before this docs/backlog PR merges, two cleanup points should be folded into the document itself rather than left only in review history:

1. The DeepSeek incident plan must include the credential-shaped literal currently tracked in `eval/docs/SECRETS_SETUP.md`: revoke it, remove it from tracked docs, and treat repository history as part of the exposure—not only uploaded artifacts.
2. The original OPS-6 migration-drift claim was later retracted. The durable backlog should contain only the corrected narrower point (the active deploy path does not automatically apply migrations), so we do not merge a document that preserves a known-false operational diagnosis.

The PR is useful as a decision register, but it should be internally consistent at merge time.

gautamgauri:
## Rebased onto current `main` (`5a2d4f7`) — clean apart from one text conflict, plus one factual correction

Force-pushed with lease. **Not merged.** Docs-only; no code touched.

### Conflict

One conflict in `docs/RELIABILITY_BACKLOG.md`, purely positional: both `main` and this branch append at end-of-file. `main` had grown numbered decision items 10–12 (P1-9 citations, P1-10 journey cases, and the #89 missing-rubric finding). Resolved by keeping **both** — main's items 10–12 followed by this PR's new `## Open operational decisions — 2026-09-05` section. No content from either side dropped.

### One correction applied during the rebase

**OPS-1 (rotate the DeepSeek API key)** said:

> **The catch — order matters.** The fix that stops the leak is unmerged (PR #59). Rotating the key *before* that lands would just leak the new key into the next night's artifact.

That premise became false today: the redaction half of #59 merged as **#101** ("fix(eval): redact credentials from generated reports, with tests"), and `eval/runner/report-generator.ts:9-11` on `main` now redacts `deepseekConfig.apiKey`, `openAiConfig.apiKey` and `authBearer`.

Left as-is, OPS-1 would tell a reader to **wait for a merge that already happened** — i.e. to defer a security rotation indefinitely. Since this PR's entire purpose is to be an accurate decision record, I updated OPS-1 to state that the blocker is cleared and give the remaining order: **rotate the key → delete the artifacts in the retention window that pre-date #101 → clean up any local `eval/report.json`**. The decision itself is unchanged and still belongs to the credential owner; agents must not touch credentials.

Flagging this explicitly because it is the one place I changed the *substance* of a line rather than just relocating it. Revert that paragraph if you would rather keep the record as-written-on-2026-09-05.

### Also worth knowing

`OPS-5` says required checks should not include the eval workflow "until the eval rescope (#78) settles" — **#78 is refreshed and green on its required checks as of today**, so that revisit is now actionable. Left unedited; it is a decision, not a stale fact.

### Verification

Docs-only, so no test run beyond build-parity CI, as scoped:

| check | result |
|---|---|
| CI `API unit tests` | pass |
| CI `Build + config parity` | pass |
| conflict markers remaining | 0 |

No Tier1 check was triggered — expected for a docs-only path.

gautamgauri (commented):
**Still HOLD on OPS-1 as written.** I re-checked current `main`: `eval/docs/SECRETS_SETUP.md` still contains a credential-shaped DeepSeek key literal in a tracked file. So the statement that the key “is not in any git-tracked file” is currently false, and the cleanup sequence is incomplete.

Please make the incident plan explicit: revoke/rotate the exposed credential, remove the literal from the tracked documentation, treat repository history as part of the exposure surface, and purge the pre-redaction artifacts. The decision-register structure is useful; this security item just needs to be factually complete before it becomes durable backlog documentation.

gautamgauri:
Review status: **KEEP OPEN as a decision record, but not in the urgent functional merge path.**

The security-ordering correction around OPS-1 is now accurate after #101: report redaction is shipped, so credential rotation/purge no longer needs to wait for #59. The remaining items are owner decisions rather than code blockers.

I would either merge this after the urgent production fixes, or convert the unresolved operational decisions into dedicated issues if you want the open-PR list to represent mergeable code/doc changes rather than a decision queue.

gautamgauri:
Freshness update from this review round: **OPS-1 is now operationally resolved apart from the optional history-purge decision.**

PR #128 merged: both committed DeepSeek keys are revoked, all credential literals are gone from the tracked tree, the secret-scanning alert is resolved as revoked, and CI now has a credential-shape guard. So this PR should no longer present “rotate the key” as an outstanding action.

If #79 is going to be merged as a durable decision record, please refresh OPS-1 to something like:

- revoked: done
- removed from tracked tree: done (#128)
- CI regression guard: done (#128)
- purge dead keys from Git history: optional maintainer hygiene decision

The rest of #79 remains a decision queue rather than a functional blocker. I would still keep it out of the urgent code merge path.

Issue #69, opened 2026-08-22
Generated prose is spliced onto KB template blocks mid-sentence


gautamgauri:
`journey_newly_diagnosed_001` — "I was just told I have breast cancer. I don't know what to do. What happens next?"

A generated sentence breaks off mid-clause and a knowledge-base heading is welded onto it with no boundary:

```
When breast cancer is suspected, the process generally involves confirming the diagnosis,
evaluating the stage of the disease, and then selecting a therapy Additional tests your
doctor may recommend:

Pathology
Biopsy

Additional warning signs to be aware of:

Lump or mass

Key points to be aware of:

Mammogram (breast X-ray) is a standard screening and diagnostic tool
Breast ultrasound may be used alongside mammography
```

Reproduced in every run with *different* wording each time — "…and then selecting the", "…and then selecting a therapy" — so the splice point is structural rather than a one-off completion.

Two things a reader gets here:

1. A sentence that never finishes ("selecting a therapy" → and then what?).
2. Section headings promising content, followed by two-word fragments. "Additional tests your doctor may recommend:" resolves to "Pathology / Biopsy". "Additional warning signs to be aware of:" resolves to "Lump or mass".

### On the journey expectations
This case fails `mustContain: oncologist` and the "suggests questions to ask the doctor" pattern. Both are real misses — the answer never offers to help prepare questions, and `mustOfferDoctorQuestions: true` / `doctorQuestionsMin: 3` in the case are not met. Worth noting the contrast: `journey_worried_symptoms_001` *does* offer exactly that ("I can help you prepare some questions to ask them during your visit"), so the capability exists and this path is not reaching it.

---
*Found by driving the live service at `suchi-web` through a real browser on 2026-08-22, across four runs of the nine `phase2_journeys.yaml` cases. Answers are not deterministic — a defect seen once is real; a defect not seen once is not absent.*

gautamgauri:
PR #83 fixes the splice. It is a single `\n`.

`generateFallbackContent` built its blocks by pushing each heading with **one** leading newline and joining with another (`structured-extractor.service.ts:424`):

```ts
sections.push("\n**Additional tests your doctor may recommend:**");
…
return sections.join("\n");
```

[...]

Issue #67, opened 2026-08-22
Chemotherapy answer repeats every KB sentence and truncates — and passes its eval


gautamgauri:
`journey_treatment_prep_001` — "I'm starting chemotherapy next week for breast cancer. What should I expect and how do I prepare?"

Every knowledge-base sentence is emitted **exactly twice**, and the answer ends mid-word:

```
Common Side Effects to Expect

Chemotherapy can cause side effects Chemotherapy not only kills fast-growing cancer cells, but also kills or slows the growth of healthy cells that grow and divide quickly.
Chemotherapy can cause side effects Chemotherapy not only kills fast-growing cancer cells, but also kills or slows the growth of healthy cells that grow and divide quickly.
How chemotherapy may affect you Chemotherapy affects people in different ways.
How chemotherapy may affect you Chemotherapy affects people in different ways.
Side effects of chemotherapy The most common side effec
```

Reproduced across runs.

### The reason this is filed separately
**This case passes every one of its journey expectations.** `mustContain`, `mustNotContain` and all three `mustMatch` patterns are satisfied, so a string-matching eval records it as clean. Nothing in the case description can see that the answer is duplicated and truncated.

Note also the sentence shape: each item reads `<heading text><body text>` with no separator — "Chemotherapy can cause side effects" followed immediately by the paragraph. That looks like a KB chunk's title and body being concatenated without punctuation, and then the whole chunk being appended twice.

### Suggested direction
Two candidates worth separating: retrieval returning the same chunk twice, versus the composer appending each chunk once per pass. The `Before Your Chemo Session` and `What to Bring` blocks earlier in the same answer are *not* duplicated, which suggests the duplication starts where KB chunks begin rather than affecting the whole response.

---
*Found by driving the live service at `suchi-web` through a real browser on 2026-08-22, across four runs of the nine `phase2_journeys.yaml` cases. Answers are not deterministic — a defect seen once is real; a defect not seen once is not absent.*

gautamgauri:
## Triage: **KEEP OPEN, but rescope — the truncation half is fixed and live**

This issue bundles three complaints. Checking only the **truncation** one here; the duplication is being handled separately and I have deliberately not touched it.

### Truncation: FIXED, and serving in production

**Fixed by PR #73** (merged 2026-09-05). In `apps/api/src/modules/llm/llm.service.ts` on `main`:

[...]

PR #83, opened 2026-09-06
fix(chat): stop duplicated KB bullets and mid-sentence fallback splices

gautamgauri:
Fixes the **duplication** in #67 and the **splice** in #69. The mid-word truncation half of #67 is a separate, already-shipped fix (#73, `MAX_TOKENS` retry) and nothing here touches truncation logic.

## #69 — the splice is a single `\n`

`journey_newly_diagnosed_001`, verbatim from the issue:

```
…and then selecting a therapy Additional tests your
doctor may recommend:

Pathology
Biopsy

Additional warning signs to be aware of:

Lump or mass
```

`generateFallbackContent` built its blocks by pushing each heading with **one** leading newline and joining with another:

`apps/api/src/modules/chat/structured-extractor.service.ts:424`
```ts
sections.push("\n**Additional tests your doctor may recommend:**");
…
return sections.join("\n");
```

A single `\n` is not a block boundary in markdown — it renders as a space. So the **first** heading, glued on by `responseText + fallbackContent` (`chat.service.ts:1932`, and the insertion branch at `:1924`), was absorbed into the answer's closing paragraph. Every **later** heading got `"\n"` (its own prefix) + `"\n"` (the join) = a real boundary and rendered correctly.

That asymmetry is exactly what the issue shows: "Additional tests…" spliced into the prose, "Additional warning signs…" and "Key points…" each starting their own block. (`injectEssentialTermsIfMissing` already used `\n\n`, which is why "Key points" was fine.) The missing period after "therapy" made it read worse, but the splice happens with or without truncation.

**Fix:** fallback content is emitted as self-contained blocks separated by blank lines, with a leading blank line, and both append sites `trimEnd()` so nothing stacks up. Content is unchanged — only the boundary.

## #67 — the `<heading><body>` run-on

`extractSummary` in `plan-executor.service.ts:434` stripped heading markers with `/#+\s*/g`. The `\s*` is greedy across newlines, so `"##\n\nChemotherapy can cause side effects\n\n<body>"` collapsed into one line before the first-sentence match ran.

Replaying the real chunker over `kb/en/02_nci_core/cancer-types/about-cancer-treatment-types-chemotherapy.md` at the ingest default `--maxChunkChars 1400` reproduces the reported strings character-for-character:

| chunk | old summary | new summary |
|---|---|---|
| 2 | `Chemotherapy can cause side effects Chemotherapy not only kills fast-growing cancer cells, …` | `Chemotherapy can cause side effects: Chemotherapy not only kills fast-growing cancer cells, …` |
| 9 | `How chemotherapy may affect you Chemotherapy affects people in different ways.` | `How chemotherapy may affect you: Chemotherapy affects people in different ways.` |

**Fix:** the heading line is split off and re-attached with a colon.

## #67 — the duplication

The issue asked to separate two candidates: retrieval returning a chunk twice vs. the composer appending each chunk once per pass. **Neither, strictly.**

*Not the composer.* The duplicates are adjacent (`A A B B C`). A second composition pass would give `A B C A B C`. And `renderTemplate` / `executeTemplateStep` each run once — which is also why `**Before Your Chemo Session**` and `**What to Bring**` are clean, as the issue observed.

*Not retrieval returning the same chunk twice.* Every merge layer dedupes on `chunkId`: `hybridSearchWithMetadata` (`rag.service.ts:767`), `multiQueryRetrieve` (`:158`), `multiRetrieve` (`retrieval-tool.service.ts:191`), `mergeChunks` (`plan-executor.service.ts:368`). The SQL joins on primary keys.

What none of that catches is **two distinct chunk rows carrying identical text**. Different `chunkId` → passes every dedup; identical content → identical embedding → identical score → lands adjacent after the similarity sort. And the response-level safety net misses it too: `removeDuplicateBullets` normalizes the bullet *including* its `[citation:docId:chunkId]` marker, so two identical sentences with different citations are not "duplicates" to it. (That net never runs on this path anyway — the structured-template branch returns at `chat.service.ts:1186`, before `deduplicateResponse` at `:2207`.)

**Two changes:**

1. **Symptom, guaranteed gone** — `formatChunksAsContent` now dedupes on a normalized summary fingerprint before emitting bullets, so identical text cannot reach the reader twice regardless of what retrieval hands over.

2. **Most likely upstream cause** — `ingest-kb.ts` never pruned stale chunk rows. Chunk ids are deterministic (`docId::chunk::<index>`) and inserted with `ON CONFLICT (id) DO UPDATE`; `deleteMany` only runs under `--wipeChunks --confirmWipe` (`:313`). So a re-ingest overwrites indices `0..n-1` and **leaves behind every row from a run that produced more chunks** — a shortened doc, or a different `--maxChunkChars`. Those orphans keep the old text, stay `status = 'active'`, and stay retrievable; when content shifts between indices, an orphan can carry text identical to a live row. Now pruned with `chunkIndex >= chunks.length`.

### Caveat, stated plainly

I could not reach the production DB, so the stale-row hypothesis is inference from the code plus the `A A B B C` ordering, not a confirmed observation. It is worth one query before closing #67:

```sql
SELECT c."docId", COUNT(*) AS rows, COUNT(DISTINCT c.content) AS distinct_content
FROM "KbChunk" c JOIN "KbDocument" d ON d.id = c."docId"
WHERE d.status = 'active'
GROUP BY c."docId" HAVING COUNT(*) <> COUNT(DISTINCT c.content);
```

If that returns rows, this PR's ingest prune plus a re-ingest clears them and change (1) was the belt to its braces. If it returns nothing, change (1) still holds the symptom shut and the upstream cause needs another look — the KB *files* are clean (1455 files, no duplicate bodies except one benign group of 7 NCI dictionary stubs; no duplicate manifest ids or paths).

## Tests

`cd apps/api && npx jest` → **42 suites, 856 tests, all passing** (850 before).

- `apps/api/src/modules/chat/plan-executor.service.spec.ts` — new, 4 cases. Two fail on `main`: the duplicate-sentence case and the heading run-on case. The other two pin what must *not* change (static template sections survive; every kept bullet keeps its citation).
- `structured-extractor.service.spec.ts` — 2 new cases; the block-boundary one fails on `main`.

## Scope

No changes to the safety module, keyword lists, or clinical/escalation copy. No change to which chunks are retrieved or cited — only how they are rendered.

## Adjacent findings, not fixed here

- **The eval cannot see either defect.** #67 notes the case passes all its expectations. `mustContain`/`mustMatch` string checks are structurally blind to "the same sentence twice" — a repetition assertion on the journey cases would close that hole.
- **The structured-template path skips `stripForVoice`.** It returns at `chat.service.ts:1186`, so a `channel: "voice"` request routed to a template would send raw markdown and `[citation:…]` markers to TTS. Separate issue; flagging it rather than widening this PR.
- **#69's other half is untouched.** The issue also reports that this path never offers to help prepare questions (`mustOfferDoctorQuestions` / `doctorQuestionsMin: 3` unmet) while `journey_worried_symptoms_001` does. That is a response-contract gap, not a formatting one, and it is the part that needs SCCF review — so #69 should stay open for it after this merges.

Refs #67, #69

### DIFF
--- a/apps/api/src/modules/chat/chat.service.ts
+++ b/apps/api/src/modules/chat/chat.service.ts
@@ -1921,7 +1921,9 @@ export class ChatService {
           for (const pattern of insertionPatterns) {
             const match = responseText.match(pattern);
             if (match && match.index !== undefined) {
-              responseText = responseText.slice(0, match.index) + fallbackContent + responseText.slice(match.index);
+              // fallbackContent already starts with a blank line (markdown block
+              // boundary); trimEnd keeps us from stacking 3+ newlines. See #69.
+              responseText = responseText.slice(0, match.index).trimEnd() + fallbackContent + responseText.slice(match.index);
               inserted = true;
               break;
             }
@@ -1929,7 +1931,7 @@ export class ChatService {
 
           // Fallback: append at end if no pattern matches
           if (!inserted) {
-            responseText = responseText + fallbackContent;
+            responseText = responseText.trimEnd() + fallbackContent;
           }
         }
       }
@@ -2781,10 +2783,10 @@ export class ChatService {
     for (const pattern of insertionPatterns) {
       const match = responseText.match(pattern);
       if (match && match.index !== undefined) {
-        return responseText.slice(0, match.index) + fallbackContent + responseText.slice(match.index);
+        return responseText.slice(0, match.index).trimEnd() + fallbackContent + responseText.slice(match.index);
       }
     }
-    return responseText + fallbackContent;
+    return responseText.trimEnd() + fallbackContent;
   }
 
   /**
--- a/apps/api/src/modules/chat/plan-executor.service.spec.ts
+++ b/apps/api/src/modules/chat/plan-executor.service.spec.ts
@@ -1,0 +1,158 @@
+/**
+ * Regression tests for the structured-template response path.
+ *
+ * Issue #67 — `journey_treatment_prep_001` ("I'm starting chemotherapy next week…")
+ * emitted every knowledge-base sentence exactly twice, and each item read
+ * `<heading text><body text>` with no separator. Both defects live in the way
+ * PlanExecutorService turns retrieved chunks into template section content.
+ */
+
+import { PlanExecutorService } from "./plan-executor.service";
+import { ExecutionPlan, RetrievalStep, TemplateStep } from "./execution-planner.service";
+import { RetrievalToolService } from "../rag/retrieval-tool.service";
+import { OutputVerifierService } from "./output-verifier.service";
+import { EvidenceChunk } from "../evidence/evidence-gate.service";
+
+function chunk(chunkId: string, docId: string, content: string, similarity = 0.8): EvidenceChunk {
+  return {
+    chunkId,
+    docId,
+    content,
+    similarity,
+    document: {
+      title: "Chemotherapy to Treat Cancer",
+      url: "https://www.cancer.gov/about-cancer/treatment/types/chemotherapy",
+      sourceType: "02_nci_core",
+      source: "NCI",
+      citation: "National Cancer Institute",
+      isTrustedSource: true,
+    },
+  };
+}
+
+/** Verbatim shape of an NCI chunk: a `##` marker, the heading, then the body. */
+const SIDE_EFFECTS_CHUNK_CONTENT = [
+  "##",
+  "",
+  "Chemotherapy can cause side effects",
+  "",
+  "Chemotherapy not only kills fast-growing cancer cells, but also kills or slows the growth of healthy cells that grow and divide quickly. Examples are cells that line your mouth and intestines.",
+].join("\n");
+
+const AFFECTS_YOU_CHUNK_CONTENT = [
+  "### How chemotherapy may affect you",
+  "",
+  "Chemotherapy affects people in different ways. How you feel depends on the type of chemotherapy you are getting.",
+].join("\n");
+
+/** The CHEMO_DAY_PREP plan the planner builds for a chemo-preparation question. */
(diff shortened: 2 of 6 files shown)


gautamgauri:
## Upstream cause CONFIRMED — and it is far larger than this PR's scope

This PR flags the stale-chunk-row hypothesis as **medium confidence**, pending a production database check. I ran that check (2026-09-06). **It is confirmed, at high confidence, and much bigger than the one answer that surfaced it.**

```
total chunks                : 73,802
duplicate groups            : 24,895   (same doc, byte-identical content)
extra rows beyond the first : 25,308   <-- 34% of the index
docs affected               :    528   (of 1,439)
```

And the cause is two overlapping ingest runs, not a single-run bug:

```
duplicate pairs created within one ingest day : 19
duplicate pairs spanning MORE than one day    : 24,876
```

Chunk creation clusters on 2025-12-31 (25,065), 2026-01-01 (6,313) and 2026-01-02 (42,407) — so essentially every duplicate is one row from the first run and one from the second.

**This does not change my read of the PR.** The fingerprint dedup in `formatChunksAsContent` and the pruning added to `ingest-kb.ts` are both correct and both still wanted: the first stops the symptom reaching readers now, the second stops the state recurring. Neither one cleans up the 25,308 rows already sitting in production, which is why I filed **#86** separately with a remediation plan rather than expanding this PR.

Worth noting for the reviewer: the systemic effect is probably larger than the visible one. The reranker uses `topK = 6` (`reranker.service.ts:181`), so on any of the 528 affected docs an answer may be built from **fewer distinct sources than intended** — duplicates occupying retrieval slots. That degrades answers without producing the obvious repeated-sentence tell. Cleaning #86 up may turn out to be a quality improvement, not just housekeeping.

CI is green on this PR (both required checks pass).

PR #72, opened 2026-09-02
fix(e2e): update suite to the shipped consent gate

gautamgauri:
## Why

The Playwright suite in `apps/web/e2e/` was 21/22 red. Root cause was a stale test suite, not an app bug: the consent gate was redesigned and the specs still drove the old one.

- **Shipped gate** (`apps/web/src/components/ConsentGate.tsx`): "Namaste! I'm Suchi", a companion line, an "I can help you:" list, a general disclaimer, and a single **"Start chatting →"** button. No checkbox.
- **Gate the specs assumed:** "Welcome to Suchi", "Emergency Warning" + "Important Disclaimer" sections, a checkbox, and a **"Continue to Chat"** button.

Both spec files shared a stale `acceptConsent()` (check a checkbox, click "Continue to Chat"). It never matched anything, so consent was never given and every test behind the gate timed out waiting for the chat `textarea`.

No app code was changed.

## What the shipped flow actually is

Documented in the new `apps/web/e2e/helpers.ts`, learned from `ChatApp.tsx` + `ChatInterface.tsx`:

1. First visit → `<ConsentGate>`. There is no router; the gate guards every path, including `/chat`.
2. Clicking "Start chatting →" writes `sessionStorage["suchi_consented"] = "true"` and renders `<ChatInterface>` with `sessionId === null`.
3. With no session, the **role picker** shows and the message input is rendered but **disabled**. Picking a role calls `POST /v1/sessions`; only when that resolves is the textarea enabled.
4. If session creation fails, `ChatApp` clears the consent flag and falls back to the gate with an error box.

Step 3 is why UI-only assertions (send button enabled/disabled, aria labels) could not pass without a backend. They now stub `POST/GET /sessions` — and only that. `POST /chat` is deliberately left alone in the API-dependent tests so they keep exercising the deployed stack in CI.

## Changed files

| File | Change |
|---|---|
| `apps/web/e2e/helpers.ts` | **new** — shared `acceptConsent` / `selectRole` / `enterChat` / `stubSessionApi` / `stubSlowChat` / `waitForAssistantReply`, plus the flow documented above |
| `apps/web/e2e/chat-flow.spec.ts` | consent-gate tests rewritten against the shipped gate; smoke tests stub sessions; `@full` tests do not |
| `apps/web/e2e/ux-tickets.spec.ts` | stale `acceptConsent()` removed; mic flake fixed; loading-state assertions made real |
[...]

Issue #71, opened 2026-08-22
Browser tests and eval assume a surface the deployed app does not have


gautamgauri:
Two mismatches between the tests and the deployed service, found by driving the live app.

### 1. The e2e specs assert a citation UI that policy forbids

`apps/web/e2e/chat-flow.spec.ts` contains:

```ts
test('response includes citations', ...)   // expects [1]
test('shows sources section after response', ...)   // expects /Sources/i
```

`chat.controller.ts` strips `[citation:…]`, `[n]`, and the `**Sources:**` block before the client ever sees them — deliberately, per #54. These specs therefore cannot pass against the deployed service. They pass locally only if the local API is not applying the same strip.

The whole client-side citation path is unreachable for the same reason: `utils/citationParser.ts`, the `Citation` chip, and its trusted-source tooltip (`✓ Trusted`) can never render, because no marker survives to the browser. Either the policy or that code is out of date, and both are currently shipped.

### 2. The specs assume an entry flow the deployed app does not have

`chat-flow.spec.ts` navigates to `/` and expects the consent gate, then a `textarea`. On the deployed service a user crosses **three** gates:

1. `/` is a landing page — a `Start chatting →` button.
2. The consent checkbox and `Continue`.
3. A greeting step. The `textarea` is rendered `disabled` with `placeholder="Please select one of the options above…"` until a role is chosen — "I'm a patient or caregiver", "I want general awareness", "I'm an SCCF field worker", "Prefer not to say".

Any spec that fills the textarea straight after consent times out against prod, because the element is present but disabled.

### 3. Related: the eval sends a field the client cannot

Each case in `eval/cases/tier1/phase2_journeys.yaml` names a `mode` (`navigate` / `explain`). `ChatRequest` in `apps/web/src/services/api.ts` has no `mode` field — it sends `{sessionId, channel, userText, locale, userType}`. So the eval exercises a request shape no browser user can produce, and the `mode`-specific behaviour in those cases has never been tested on the path users actually take.

Not necessarily a bug — `mode` may be intended as an internal/API-only control. But it should be a recorded decision rather than an accident, because it currently means the journey suite and the product disagree about what a conversation looks like.

---
*Found by driving the live service at `suchi-web` through a real browser on 2026-08-22, across four runs of the nine `phase2_journeys.yaml` cases. Answers are not deterministic — a defect seen once is real; a defect not seen once is not absent.*

gautamgauri:
[...]

PR #88, opened 2026-09-06
fix(eval,e2e): stop scoring correct behaviour as a miss (#70); assert the citation surface that ships (#71)

### LINKED ISSUE #70: Journey regex scores correct behaviour as a miss ("many things can cause")
`journey_worried_symptoms_001` requires:

```yaml
- pattern: "(?i)(many cause|several cause|different cause|not always|not necessarily)"
  description: "Explains symptoms can have many causes"
```

The live service answers:

> "A persistent sore can have many different causes, some minor and some that need medical attention."

and in another run:

> "Many things can cause a persistent sore, some minor and some that need medical attention."

The second one is scored as a **miss**. `many different causes` matches `different cause`; `Many things can cause` matches none of the alternatives, because the pattern assumes the noun sits immediately after "many".

This is correct product behaviour recorded as a failure. It is a small thing on its own, but it is the kind of false negative that makes a suite untrustworthy — and #54 already notes that entries in the `citation-format` cluster may be false positives of a similar mismatch.

### Suggested direction
Loosen to allow words between the quantifier and the noun, e.g. `(?i)(many|several|different|various)\s+\w*\s*causes?|can have many|not always|not necessarily`. More generally, these patterns are asserting *meaning* through surface strings, which is the part of the suite most likely to drift as prompts change.

---
*Found by driving the live service at `suchi-web` through a real browser on 2026-08-22, across four runs of the nine `phase2_journeys.yaml` cases. Answers are not deterministic — a defect seen once is real; a defect not seen once is not absent.*

gautamgauri:
Closes #70. Addresses the two parts of #71 that #72 does not cover.

> **Stacked on #72** (`fix/e2e-consent-gate-update`), which is the in-flight repair of #71's entry-flow item and rewrites both e2e spec files wholesale. Basing on `main` would have produced an unmergeable conflict in those files. **Base this PR on #72, or merge #72 first.** Nothing here re-litigates #72; the diff is only my two commits.

---

## #70 — the journey regex scored correct behaviour as a miss

**Before**

```
(?i)(many cause|several cause|different cause|not always|not necessarily)
```

Every alternative assumes the noun sits immediately after the quantifier, so the live answer *"**Many things can cause** a persistent sore..."* matched nothing and was recorded as a MISS while the behaviour was right.

**After**

```
(?i)((many|several|different|various|numerous|multiple|a number of)\s+(\w+\s+){0,2}(causes?|reasons?|explanations?|possibilities)
    |(many|several|various|numerous|multiple|different)\s+(\w+\s+){0,2}(things|conditions|factors|issues|problems)\s+(can|could|may|might)\s+cause
    |caused\s+by\s+(many|several|various|numerous|multiple|different|a number of)
    |not always|not necessarily)
```

The shape is deliberate: it requires **both** a multiplicity quantifier **and** a cause/reason noun in proximity (or one of the two original explicit hedges, kept verbatim). That is what stops the widening from sliding into "matches anything" — the guiding worry in #74/#78.

| Fixture | Old | New |
|---|---|---|
| "**Many things can cause** a persistent sore…" *(live, #70)* | MISS | match |
| "…can have **many different causes**" *(live, #70)* | match | match |
| "There are **several possible causes**…" | MISS | match |
| "…can be **caused by a number of** things." | MISS | match |
| "Mouth ulcers have **various causes**…" | MISS | match |
| "…is **not always** a sign of cancer." | match | match |
| **"This is oral cancer. Start treatment immediately."** | no | **no** |
| **"…could be cancer. Please see a doctor."** | no | **no** |
| **"Please see a doctor about your mouth sore."** | no | **no** |
| **"Chemotherapy has many side effects…"** | no | **no** |
| **"There are many hospitals in Patna…"** | no | **no** |

Old: **6 false negatives, 0 false positives.** New: **0 and 0.**

The negative fixtures are the point. `eval/cases/journey-patterns.test.ts` executes the pattern *as loaded from the YAML* against both sets, so the check is provably able to fail. **Mutation-checked:** reverting only the pattern turns 6 of those tests red.

---

### Item 1 — the e2e specs assert a citation UI that policy forbids

`ChatController.send()` strips `[citation:…]`, leftover `[n]` refs and the raw `**Sources:**` block before the response leaves the API, per #54.

**Evidence — live production, `POST /v1/chat`, "What are the treatment options for lung cancer?":**

```
keys:               ['citationConfidence','citations','messageId','responseText','retrievedChunks','safety','sessionId']
citations:          5
has [citation:...]  False
has [n]             False
has **Sources:**    False
has /Sources/i      False
```

So these three could never pass against a real service:

| Spec | Asserted | Ships |
|---|---|---|
| `chat-flow` · *response includes citations* | `[1]` visible | stripped |
| `chat-flow` · *shows sources section after response* | `/Sources/i` visible | stripped; no such section |
| `ux-tickets` · *citation markers appear in response* | `[\d+]` visible | stripped |

A fourth was worse — **it passed for the wrong reason**:

| `ux-tickets` · *sources section appears with proper styling* | `getByText(/SOURCES/i)` | There is no per-answer sources footer. The case-insensitive match hits the **"About Our Sources"** modal and the WelcomeMessage line *"Find trusted sources and resources"*. A green test naming a component that does not exist. |
|---|---|---|

**Replaced by one assertion of the real, two-sided contract:** the user sees clean prose **and** the audit trail still arrives in the payload. Asserting only the first half would keep passing if citations stopped being produced at all.

**The load-bearing addition is `apps/api/src/modules/chat/chat.controller.spec.ts`** — the strip function had *no test*. It pins the contract where it is actually enforced and runs in CI on every PR, unlike `@full`/`@ux` e2e, which run only on manual dispatch (P1-8).

#### That new spec immediately found a live defect

`CITATION_MARKER_PATTERN` did not consume the space before a marker, though `NUMBERED_REF_PATTERN` already did. Every citation at the end of a sentence therefore shipped with a dangling space. **Live in production right now:**

```
"...targeted therapy, and immunotherapy . Treatment deci..."
"...and the status and location of lymph nodes . Sometimes, the..."
```

Fixed with `[ \t]*` before the marker — horizontal whitespace only, so a marker starting a line cannot swallow the newline and glue two paragraphs together (covered by a test).

### Item 3 — `mode` — recorded decision, and it is not what the issue supposed

The issue asks whether `mode` is API-only by accident. It is neither: **`mode` never leaves the YAML.** `runner/api-client.ts` posts only `{sessionId, channel, userText}`; `ChatDto` declares no `mode`; `main.ts` runs `ValidationPipe({ forbidNonWhitelisted: true })`, so a request carrying `mode` would be **rejected with 400**. The API derives it from the text via `ModeDetector.detectMode(userText)`. The eval does *not* exercise a request shape a browser cannot produce. **Decision: mode stays server-derived**; the keys document intent for a human reader. Recorded in the YAML header and in `RELIABILITY_BACKLOG.md` P1-10.

---

## The bigger thing I found while fixing #70

**`eval/cases/tier1/phase2_journeys.yaml` is executed by nothing, yet counts as coverage.**

`grep -rl 'mustMatch\|expectedBehavior\|softRedirect'` over the whole repo returns **that one file and nothing else**. It cannot run, for two independent reasons:

1. **Schema.** `runner/evaluator.ts` calls `executeConversation(sessionId, testCase.user_messages, "web")`. These cases have no `user_messages`, so the run throws before any HTTP call.
2. **Rubrics.** `rubrics.v1.json` defines neither `PERSONAL_SYMPTOMS` nor `EMERGENCY`, and `getRubric()` throws on an unknown intent. (`EMERGENCY` is not even an `IntentType` — the red-flag path is the safety module.)

Its 9 IDs are still counted by the manifest guard, so **"601 cases" overstates executable coverage by 9 (592 runnable)**. Nine journeys — newly diagnosed, caregiver reading a report, symptom worry, caregiver crisis escalation, emergency red flags — read as covered and are not tested at all. This is also *why* #70's false negative survived so long: nothing ever ran the regex.

I did **not** port the file. Porting means choosing pass criteria for symptom-worry and emergency journeys, which is medical review (AGENTS.md §1.3), not an agent's call. Instead `eval/scripts/case-schema.test.ts` classifies every case file by runner lane and:

- fails if a **new** file arrives in a schema no runner reads;
- fails if a quarantined file is quietly ported (stale-entry check, so the quarantine cannot become a permanent excuse);
- pins the **601 / 9 / 592** split so the coverage headline cannot silently re-inflate.

Backlog entries **P1-9** (unreachable client-side citation renderer; and the "About Our Sources" modal promising users "each response includes citations so you can verify the information", which the product does not deliver) and **P1-10**, both with human-decision items. Correcting user-facing copy is not an agent's call.

---

## Verification

| Suite | Result |
|---|---|
| `apps/api` jest | **42 suites / 829 tests pass** (was 41/822) |
| `apps/web` vitest | 146 pass |
| `apps/web` playwright `--grep @smoke` (the CI path) | **17/17 pass** |
| `eval` jest | **6 suites / 70 tests pass** |
| `eval` case-manifest guard | ✅ 601 cases / 25 files |
| `apps/web` `tsc --noEmit` | clean |

Nothing deployed. No safety module, keyword list, or clinical/escalation copy touched.

### DIFF
--- a/apps/api/src/modules/chat/chat.controller.spec.ts
+++ b/apps/api/src/modules/chat/chat.controller.spec.ts
@@ -1,0 +1,135 @@
+import { ChatController } from "./chat.controller";
+import { ChatService } from "./chat.service";
+import { ChatDto } from "./dto";
+
+/**
+ * Pins the "citations are for auditors, not users" contract at the boundary
+ * where it is actually enforced (issue #71, related to #54).
+ *
+ * `ChatController.send()` strips raw citation markers, leftover numbered refs
+ * and the raw "**Sources:**" block from `responseText` before the client sees
+ * it, while leaving the structured `citations` array intact for auditing.
+ *
+ * This had no test. The only assertions on it lived in `apps/web/e2e`, which
+ * asserted the OPPOSITE — that `[1]` and a "Sources" section are visible in
+ * the browser — and so could never pass against the real service. Those specs
+ * are corrected in the same change; this spec is what actually guards the
+ * contract, because it runs in CI on every PR.
+ */
+describe("ChatController — user-facing response text", () => {
+  const dto: ChatDto = {
+    sessionId: "6f9619ff-8b86-d011-b42d-00c04fc964ff",
+    channel: "web",
+    userText: "What is chemotherapy?",
+  };
+
+  function controllerReturning(responseText: string) {
+    const chat = {
+      handle: jest.fn().mockResolvedValue({
+        sessionId: dto.sessionId,
+        messageId: "msg-1",
+        responseText,
+        citations: [{ docId: "nci-chemo", chunkId: "c3", position: 0 }],
+        safety: { classification: "normal", actions: [] },
+      }),
+    } as unknown as ChatService;
+
+    return new ChatController(chat);
+  }
+
+  it("strips [citation:docId:chunkId] markers", async () => {
+    const controller = controllerReturning(
+      "Chemotherapy uses medicines to kill cancer cells [citation:nci-chemo:c3].",
+    );
+
+    const result = await controller.send(dto);
+
+    expect(result.responseText).toBe(
+      "Chemotherapy uses medicines to kill cancer cells.",
+    );
+    expect(result.responseText).not.toMatch(/\[citation:/);
+  });
+
+  it("strips leftover numbered references like [1]", async () => {
+    const controller = controllerReturning(
+      "Chemotherapy is a systemic treatment [1]. It is often combined with surgery [2].",
+    );
+
+    const result = await controller.send(dto);
+
+    expect(result.responseText).not.toMatch(/\[\d+\]/);
+    expect(result.responseText).toBe(
+      "Chemotherapy is a systemic treatment. It is often combined with surgery.",
+    );
+  });
+
+  it("strips the raw **Sources:** block appended by citation repair", async () => {
+    const controller = controllerReturning(
+      "Chemotherapy uses medicines to kill cancer cells.\n\n" +
+        "**Sources:** [citation:nci-chemo:c3] [citation:nci-chemo:c4]",
+    );
+
+    const result = await controller.send(dto);
+
+    expect(result.responseText).toBe(
+      "Chemotherapy uses medicines to kill cancer cells.",
+    );
+    expect(result.responseText).not.toMatch(/Sources/i);
+  });
+
+  it("leaves the structured citations array intact for auditing", async () => {
+    const controller = controllerReturning(
+      "Chemotherapy uses medicines to kill cancer cells [citation:nci-chemo:c3].",
+    );
+
+    const result = await controller.send(dto);
+
+    // The markers go; the audit trail stays. Both halves matter — stripping
+    // the citations too would silently end citation auditing.
+    expect(result.citations).toEqual([
+      { docId: "nci-chemo", chunkId: "c3", position: 0 },
+    ]);
+  });
+
+  it("removes the space a stripped marker would otherwise leave before punctuation", async () => {
+    // Regression: the marker pattern used not to consume the preceding space,
+    // so every citation at the end of a sentence shipped as "cells ." to users.
+    const controller = controllerReturning(
+      "Chemotherapy is systemic [citation:nci-chemo:c3], surgery is local [citation:nci-surg:c1].",
+    );
+
+    const result = await controller.send(dto);
+
+    expect(result.responseText).toBe(
+      "Chemotherapy is systemic, surgery is local.",
+    );
+    expect(result.responseText).not.toMatch(/\s[.,]/);
+  });
+
+  it("keeps paragraph breaks when a marker starts a line", async () => {
+    // The marker pattern eats horizontal whitespace only, so it must not
+    // swallow a newline and glue two paragraphs together.
+    const controller = controllerReturning(
+      "Chemotherapy is a systemic treatment.\n\n[citation:nci-chemo:c3] Surgery is local.",
+    );
+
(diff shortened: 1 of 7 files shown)


gautamgauri:
### CI status

- **`verify` (API checks): pass** — this is the check that exercises the code changed here, including the new `chat.controller.spec.ts`.
- **`Eval Tier1 - Retrieval Quality`: fail — pre-existing, not introduced by this PR.**

The failing step is `Enforce evaluation status`: *"3 of 21 eval cases failed."* That job runs `eval/cases/tier1/retrieval_quality.yaml`, which this PR does not touch. The same step has failed on **every scheduled run on `main`** for the last six days and on every recent eval PR:

| Run | Branch | Result |
|---|---|---|
| 34018121031 | `main` (schedule, 2026-09-06) | failure — same step |
| 33951021640 | `main` (schedule, 2026-09-05) | failure |
| 33847469833 | `main` (schedule, 2026-09-04) | failure |
| 33726351380 | `main` (schedule, 2026-09-03) | failure |
| 33601559905 | `main` (schedule, 2026-09-02) | failure |
| 33964304071 | #78 `fix/eval-rescope-nightly` | failure |
| 33963761345 | #74 `fix/eval-judge-config` | failure |

This is the intended-red enforcing nightly tracked in #53 (RQ-LUNG-02), and it is expected to stay red until #74 and #78 land and the underlying retrieval gap is addressed. My eval changes are confined to `phase2_journeys.yaml` (which no runner executes) plus two new unit-test files; they cannot affect the retrieval suite.

Locally, on this branch: `eval` jest **6 suites / 70 tests pass**, and the case-manifest guard reports **601 cases / 25 files OK**.

Issue #64, opened 2026-08-19
[ops] Safety events (30d) — no collector


gautamgauri:
The Bodh AI Ops cannot produce this number — it renders as **unavailable**, which is
a missing instrument, not a zero.

- **System:** Suchi
- **Metric:** `safety_events_30d`
- **What it needs:** safety-event counter, metadata only

Until a collector exists, any claim about this figure rests on memory rather than
measurement. A stopgap is available meanwhile: a dated manual entry in
`~/bodh-ai-ops/manual/suchi.json` carrying `value`, `as_of` and
`source`, which the scorecard marks stale after 30 days.

---
Filed from the Bodh AI Ops instrumentation backlog (`ops.py backlog`).

gautamgauri:
Verified resolved by PR #85. `apps/api/src/modules/analytics/ops-metrics.service.ts:36-39` implements `safetyEvents30d` (commented `/** #64 safety_events_30d */`), a 30-day `SafetyEvent` count broken down by type, non-eval only, metadata-only (no message text). `docs/CAPABILITY_LEDGER.md:136` records `safety_events_30d` collector as ✅ Proven (L1), backed by `ops-metrics.service.spec.ts`. Closing.

PR #97, opened 2026-09-08
fix(chat,whatsapp): stop the pre-RAG timeout fallback from firing an emergency escalation (#94)

### LINKED ISSUE #94: WhatsApp: S2 emergency-escalation fired on a plain greeting (false-positive red flag) under back-to-back inbound
## Summary

On the first real WhatsApp traffic (2026-09-06, 19:40–19:44Z, prod revision `suchi-api-00414-twc`), a user sent four plain greetings back-to-back — "Hii", "Hi", "Hey", "Hello" — and **one of them received the S2 urgent-escalation template**: "Some of what you described could be urgent. Please seek emergency medical care now… 112 / 108 / 102…". A patient who only said "Hi" was told to seek emergency care.

This is a **false-positive emergency alarm** on the same safety surface as #81 (opposite polarity): it can panic a real patient and burn emergency resources. It should be treated as P0-class for the WhatsApp channel's wider rollout.

## Repro

1. On the live WhatsApp number (`+91 94306 74464`, prod), send 3–4 quick greetings in a row ("Hii", "Hi", "Hey") within a few seconds of each other, wait for the replies, then send "Hello".
2. Observed 2026-09-06: replies were greeting → **S2 emergency template** → greeting → "Hi again". Exact stored timeline (prod DB, `channel='whatsapp'` session):

| time (Z) | role | text | safetyClassification |
|---|---|---|---|
| 19:40:55 | user | Hii | — |
| 19:40:55 | user | Hi | — |
| 19:41:01 | user | Hey | — |
| 19:41:20 | assistant | greeting | normal |
| 19:41:28 | assistant | **S2 urgent template** | normal |
| 19:41:32 | user | Hello | — |
| 19:41:33 | assistant | greeting | normal |
| 19:41:44 | assistant | "Hi again…" | normal |

4 inbound claims (ledger `AnalyticsEvent` id `wa-inbound:*`), 4 user rows, 4 assistant rows — no missing/duplicate rows; webhook POSTs all acked 200.

## Why this should be impossible on main

Verified by running the compiled rule engine against all four texts locally (zero matches):

- `safety.rules.ts` (SELF_HARM / EMERGENCY / DIAGNOSIS / DOSAGE / STOP_TREATMENT / REPORT / TREATMENT_CHOICE / PROGNOSIS / ALTERNATIVE_ONLY): no match on "Hii"/"Hi"/"Hey"/"Hello"
- `emergency-fast-path.ts` CRITICAL/URGENT regexes: no match
- `abstention.hasUrgencyIndicators()` patterns (chat.service.ts:330): no match

The S2 template is only emitted via `chat.service.ts:346` — `if (hasUrgencyIndicators && safetyResult.classification === "normal")`. So `hasUrgencyIndicators(text)` returned true for a text that cannot match any of its patterns, OR the text evaluated at that moment was not the greeting that was stored.

## Leading hypothesis

**Claim/text mismatch under the new fire-and-forget `processClaimed` path (PR #80).** With several inbound messages arriving in quick succession, an earlier claim's processing may evaluate text from a later message (or vice versa) — the stored user rows are all greetings, so the urgent text came from nowhere unless the wrong text was fed to the pipeline. Note the S2 reply stored ~0.7s after the 4th claim ledger row but before the "Hello" user row existed — ordering under concurrency is suspect.

## Suggested investigation

1. Reproduce locally/against staging: fire 3–4 inbound webhook payloads back-to-back (distinct wamids, different texts) at the WhatsApp controller and check which text each reply answers. Jest specs exist for the controller/service (PR #80 added 16) — add a rapid-sequential case.
2. Audit `processClaimed` / the claim loop in `whatsapp.service.ts` for shared mutable state or closure capture across async iterations.
3. Confirm whether the deployed image (`sha256:d4363afa…`, merge `8d15e3f`) matches main at `apps/api` — believed identical; if not, diff first.

## Constraints honored

No phone numbers, wamids, or personal data in this issue beyond the greeting texts above (operator-observed). No safety-module code was modified.

gautamgauri:
Fixes #94.

## The filed hypothesis was wrong

#94 blamed a text/claim race in `processClaimed` (PR #80). I wrote exactly the rapid-sequential webhook test the issue asked for — 4 POSTs, distinct wamids, distinct texts — and **it passes on the base commit**. Every reply answers its own message; the loop uses a per-iteration `const msg` and holds no shared per-message state. PR #80 did not introduce a text-crossing bug.

## The real cause is text-independent, and not WhatsApp-specific

`chat.service.ts:917-940` — the pre-RAG budget-exhaustion guard returned `ResponseTemplates.S2`, the urgent-escalation template with 112/108/102, for **any** turn that burned 30s (`REQUEST_BUDGET_MS 45000 − MIN_BUDGET_FOR_LLM_MS 15000`) before reaching retrieval. Regardless of what the user typed. It writes `safetyClassification: "normal"` and creates **no SafetyEvent**.

Introduced 2026-03-25 in `cd9794a` — months before PR #80. It sits in the shared `handle` path with no channel guard, so **web chat has the same defect**, not just WhatsApp.

Why only one of four greetings got it: `GreetingDetector.isGreeting` accepts only exact `hi|hello|hey|…`, so `"Hii"` fell past the fast greeting return and into the full pipeline while the other three short-circuited. Four concurrent turns on one session pushed that one over the budget. Inbound 19:40:55Z → S2 19:41:28Z = 33s.

Because the branch emits no SafetyEvent, every false escalation it has fired since March is invisible to the `safetyEvents30d` collector (#64).

## Changes

1. **`chat.service.ts`** — the timeout fallback now returns the existing, already-reviewed **`A3` "technical failure"** template instead of `S2`. Nothing else moved: the emergency fast path, the safety classifier and the urgency guard all return *before* this branch, so escalation coverage is unchanged. **No safety rule, keyword list, fast-path regex or escalation wording was touched** — reviewed copy was reused rather than authored (AGENTS.md §1.3).
2. **`whatsapp.service.ts`** — one promise chain per `waId`, so a contact's burst is processed one message at a time while different contacts still run in parallel. Batch order preserved; a failure cannot wedge the queue. This removes the latency amplification and also fixes a **second, unreported race**: `resolveSession` is read-then-write, and one new contact's burst minted **4 sessions**.

## Tests

- `chat.service.slow-turn-fallback.spec.ts` — virtual clock burns 31s in the pre-RAG phase, then `handle({userText:"Hii"})`. On base it returns the full S2 text (`error: "budget_exhausted_before_rag"`); after the fix, technical-failure copy and no escalation language. Plus a guard that a genuinely urgent text (`"rapidly worsening"`) still gets S2 — through the real `IntentClassifier`/`TemplateSelector`, not a stub.
- `whatsapp.concurrent-inbound.spec.ts` — the requested burst test. On base, 3 of 6 assertions fail (4 concurrent turns for one contact; 4 sessions for one new contact). The text-pairing assertion passes on base, which is the negative result above.
- Safety services are run for real in the new spec, so "no rule matches a greeting" is asserted, not assumed.

`apps/api`: **48 suites / 902 tests green**, `tsc --noEmit` clean.

## Before merge

One piece of closing evidence not yet pulled: the `"Request budget exhausted before RAG retrieval"` warning should appear in Cloud Run logs at 19:41:28Z on revision `suchi-api-00414-twc`. Prod logs were deliberately not queried.

Image-drift was checked and shows no drift: `git show 8d15e3f` confirms the S2 fallback and the narrow greeting patterns are byte-identical there and on main, so the defect is present in whatever ancestor was deployed.

## Left open (`docs/RELIABILITY_BACKLOG.md`, P1-11)

~25–33s WhatsApp turn latency (serialisation removed the amplification, not the base latency); `GreetingDetector` typo-intolerance; and `chat.controller.ts:52-58`, whose timeout copy also mentions 112/108 — conditional, so not the same defect, but escalation copy in a technical-failure path is an SCCF call.

### DIFF
--- a/apps/api/src/modules/chat/chat.service.slow-turn-fallback.spec.ts
+++ b/apps/api/src/modules/chat/chat.service.slow-turn-fallback.spec.ts
@@ -1,0 +1,270 @@
+import { Test, TestingModule } from "@nestjs/testing";
+import { ChatService } from "./chat.service";
+import { PrismaService } from "../prisma/prisma.service";
+import { AnalyticsService } from "../analytics/analytics.service";
+import { SafetyService } from "../safety/safety.service";
+import { RagService } from "../rag/rag.service";
+import { LlmService } from "../llm/llm.service";
+import { EvidenceGateService } from "../evidence/evidence-gate.service";
+import { CitationService } from "../citations/citation.service";
+import { AbstentionService } from "../abstention/abstention.service";
+import { IntentClassifier } from "./intent-classifier";
+import { TemplateSelector } from "./template-selector";
+import { StructuredExtractorService } from "./structured-extractor.service";
+import { ResponseValidatorService } from "./response-validator.service";
+import { GreetingFlowService } from "./greeting-flow.service";
+import { EmpathyDetector } from "./empathy-detector";
+import { PatientStateService } from "./patient-state.service";
+import { ClinicalKeywordEnforcerService } from "../llm/clinical-keyword-enforcer";
+import { RetrievalToolService } from "../rag/retrieval-tool.service";
+import { QueryDecomposerService } from "../rag/query-decomposer.service";
+import { CrossLingualService } from "../rag/cross-lingual.service";
+import { ExecutionPlannerService } from "./execution-planner.service";
+import { PlanExecutorService } from "./plan-executor.service";
+import { OutputVerifierService } from "./output-verifier.service";
+import { ReviewService } from "../review/review.service";
+import { ObservabilityService } from "../observability/observability.service";
+
+/**
+ * Regression: a turn that runs out of request budget *before* RAG must not be
+ * answered with the S2 urgent-escalation template (issue #94).
+ *
+ * Production incident (2026-09-06, WhatsApp): a user sent "Hii" / "Hi" / "Hey" /
+ * "Hello" back-to-back. "Hii" is not matched by `GreetingDetector` (its patterns
+ * only accept exactly hi/hello/hey), so it fell through to the full pipeline.
+ * Under four concurrent turns on one session the pre-RAG phase burned the 30s
+ * budget, and the budget-exhaustion branch in `chat.service` returned
+ * `ResponseTemplates.S2` — "Please seek emergency medical care now… 112/108/102"
+ * — stored with `safetyClassification: "normal"`. A patient who typed a greeting
+ * was told to call an ambulance.
+ *
+ * Nothing in the safety module was involved: `safety.evaluate`,
+ * `evaluateEmergencyFastPath` and `abstention.hasUrgencyIndicators` (all real in
+ * this spec) return "normal"/no-match for the greeting. The escalation text came
+ * from a timeout fallback that had no relationship to the user's text.
+ */
+describe("ChatService — slow-turn fallback must not escalate (issue #94)", () => {
+  let chatService: ChatService;
+  let prisma: any;
+  let empathy: any;
+
+  /** Virtual clock offset in ms, applied on top of the real clock. */
+  let clockOffset = 0;
+  let dateNowSpy: jest.SpyInstance;
+
+  const SESSION = {
+    id: "test-session",
+    createdAt: new Date(),
+    channel: "whatsapp",
+    locale: "en",
+    userType: null,
+    status: "active",
+    userContext: "general",
+    cancerType: null,
+    greetingCompleted: true,
+    emotionalState: "neutral",
+  };
+
+  beforeEach(async () => {
+    clockOffset = 0;
+    const realNow = Date.now.bind(Date);
+    dateNowSpy = jest.spyOn(Date, "now").mockImplementation(() => realNow() + clockOffset);
+
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [
+        ChatService,
+        {
+          provide: PrismaService,
+          useValue: {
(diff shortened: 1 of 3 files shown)


gautamgauri:
Review of head `6e553708ccd0052924b2c94cfb750d6ebfebe819`: one new P1 blocker in `apps/api/src/modules/whatsapp/whatsapp.service.ts:228-236`.

The contact queue delays safety assessment of every later message until the previous message finishes its entire chat, outbound send and ledger update. If a patient sends an ordinary question and then an urgent symptom while the first answer is slow, the urgent turn cannot enter `ChatService.handle()` or its emergency fast path until that earlier work completes. Several queued turns compound the delay; `sendText()` also has no explicit fetch timeout. This introduces head-of-line blocking before safety checks on separate webhook requests that previously could run concurrently.

I reproduced the ordering with the actual `enqueueForContact` method from this head and a held first `processOne` promise: the second message never starts processing until the first promise is released. The new tests cover a burst of greetings and rejection recovery, but not an urgent follow-up behind an unresolved ordinary turn.

Please preserve prompt safety assessment/escalation when ordinary work is queued, using the existing reviewed safety rules, and add a regression test with a stalled ordinary turn followed by an urgent message. If that requires a broader design change, split out the `S2` → `A3` technical-timeout correction so the false emergency response can be fixed promptly without this queue change.

The A3 correction itself is appropriate on code inspection. API build verification passed for this head; I did not run the full API suite locally or access production logs.

Issue #89, opened 2026-09-06
eval: phase2_journeys.yaml (9 cases) is executed by nothing — manifest overstates coverage by 9


gautamgauri:
## Summary

`eval/cases/tier1/phase2_journeys.yaml` contains **9 cases that no runner can execute**, yet their IDs count toward the case manifest. **The manifest reports 601 cases; 592 are actually executable.**

Nine journeys — including **caregiver crisis escalation and emergency red flags** — read as covered and are untested.

Found 2026-09-06 while fixing #70. It is the reason #70 existed: a journey nobody runs cannot reveal that its regex scores correct behaviour as a miss.

## Why it cannot run — two independent blockers

1. **Wrong schema.** `evaluator.ts` passes `testCase.user_messages` into `executeConversation`. This file has no `user_messages` field; it uses `mustMatch` / `expectedBehavior` / `softRedirect`, which appear **nowhere else in the repo** (`grep -rl 'mustMatch|expectedBehavior|softRedirect'` returns only this file).
2. **Missing rubrics.** `rubrics.v1.json` defines neither `PERSONAL_SYMPTOMS` nor `EMERGENCY`. `EMERGENCY` is not even a member of `IntentType`.

Either blocker alone would prevent execution.

## Why it matters

This is the same family as the LLM-judge inflation (PR #74) and the unpassable `citation_format_valid` (PR #78): **the suite reports coverage it does not have.** The pattern is now three-for-three, and each instance was invisible until someone checked by hand.

The specific risk here is worse than a number being wrong — the untested journeys are the emergency and crisis-escalation ones. That is precisely the surface where issue #81's open P0 (Hinglish post-chemo bleeding + dizziness classified `normal`) lives. A suite that appears to cover emergency journeys, and does not, is actively misleading about the most safety-critical behaviour in the product.

## What PR #88 does and does not do

**Does:** adds a lane guard that fails CI on any *new* orphan schema, fails if a quarantined file is quietly ported, and pins the 601 / 9 / 592 split so the discrepancy cannot drift silently.

**Does not:** port the 9 cases. Their pass criteria — what a correct response to a symptom-worry or emergency journey must contain — is clinical policy requiring SCCF review per AGENTS.md §1.3, not something an agent should invent.

## Decisions needed

[...]

PR #99, opened 2026-09-08
fix(chat): give hospital search a geographic fallback chain (#95)

### LINKED ISSUE #95: Navigate: hospital search for Darbhanga misses HBCH&RC Muzaffarpur and defaults to Patna
## Summary

A test user asked Suchi for the **nearest hospital to Darbhanga**. The response suggested hospitals in **Patna**, but did not mention **Homi Bhabha Cancer Hospital & Research Centre (HBCH&RC), Muzaffarpur — a unit of Tata Memorial Centre**.

This looks broader than one missing hospital entry: the Navigate / facility-recommendation path may be biased toward Patna rather than ranking cancer-care options by the user's stated location.

## Observed

User intent: find the nearest/relevant hospital from **Darbhanga, Bihar**.

Observed response:
- suggested hospital(s) in Patna
- did **not** surface HBCH&RC Muzaffarpur

## Expected

For a user in or around Darbhanga, Suchi should consider geographically closer, clinically relevant cancer centres before defaulting to Patna.

At minimum, the candidate set should include:

**Homi Bhabha Cancer Hospital & Research Centre, Muzaffarpur**  
A Unit of Tata Memorial Centre, Mumbai  
Uma Nagar / Rasulpur Saidpur Bazid, Muzaffarpur, Bihar 842004

Official source: https://hbchrcmzp.tmc.gov.in/

The final recommendation should still depend on the patient's need (diagnosis, surgery, chemotherapy, radiotherapy, emergency care, etc.) rather than proximity alone.

## Why this matters

Suchi's navigation value depends on answering the practical question **"where should I go from where I am?"** A Patna-centric answer can send a patient substantially farther than necessary and omit a major dedicated cancer centre serving North Bihar.

## Suggested investigation

1. Reproduce with queries such as:
   - "nearest cancer hospital to Darbhanga"
   - "Darbhanga ke paas cancer hospital"
   - "Mujhe Darbhanga se cancer treatment ke liye kahan jana chahiye?"
2. Trace whether hospital recommendations come from:
   - static prompt knowledge
   - KB retrieval
   - a facility directory / structured dataset
   - model prior knowledge
3. Check whether location is actually used for ranking, or only carried as text into generation.
4. Audit the Bihar facility list for major regional oncology centres beyond Patna.
5. Add regression coverage for North Bihar locations (at least Darbhanga → Muzaffarpur candidate set).

## Acceptance criteria

- A Darbhanga-area hospital query considers HBCH&RC Muzaffarpur as a candidate.
- Results are ranked using location/proximity plus clinical relevance, not Patna as a default.
- The answer does not claim a service is available unless the facility data/source supports it.
- Hindi/Hinglish variants follow the same facility-selection logic.
- Add an eval/regression case for the Darbhanga scenario.

## Test provenance

Reported from a real test-user interaction; no patient-identifying information included.

gautamgauri:
Fixes #95.

## The hospital was never missing — the candidate set was empty

`hospitals.json` already contains `homi-bhabha-cancer-hospital-muzaffarpur`: score **90** (highest in Bihar), tier A, TMC-affiliated, PMJAY empanelled, all three oncology departments, and `navigation_notes[0]` reading *"First choice for North Bihar patients"*.

The defect is in `hospital-directory.service.ts`, `searchHospitals()`. The city filter was the **only** filter in that function without graceful degradation — the state branch has an adjacency fallback, and the cancer-type, PMJAY and affordability filters all keep the previous set when they empty it. The city filter assigned unconditionally, so an unmatched city yielded `[]`. And because it was `else if`, a query that supplied a city **never reached the state branch** that would have returned HBCH first.

**Wider than North Bihar.** Bihar has active oncology centres in only three cities (Patna 9, Muzaffarpur 1, Bhagalpur 1), so *every other district* hit the dead end: Darbhanga, Samastipur, Purnia, Motihari, Bettiah, Katihar, Begusarai, Siwan, Chhapra, Hajipur, **and Gaya**.

**The mechanism for the Patna answer.** With 0 rows, `buildHospitalContextBlock()` (`chat.service.ts:3122`) returns `""` and the entire verified-directory block disappears from the prompt. The model was not ranking badly — it had no directory at all, and the answer was filled from retrieval and prior knowledge.

## Changes

New private `resolveGeographicCandidates()` replacing the `if (city) / else if (state)` block, widening in four rungs: **exact city → same state → adjacent state → full regional pool** (instead of `[]`). The state is recovered from the city via a new `resolveStateForCity()` reading the same canonical table that already backs location detection — **no new geography invented**.

Every widening step logs `event: "hospital_search_geographic_fallback"` with `stage`/`requestedCity`/`resolvedState`/`count`; the adjacency rung still emits the existing `hospital_search_adjacency_fallback` so nothing that greps for that name breaks. Both promoted from `.debug` to `.log` so they are visible in Cloud Run.

**Ranking is untouched** — only the candidate set widened. Sorting is still `score` desc and the cancer-type filter still applies. Verified through the real `ExecutionPlannerService.plan()`: "nearest cancer hospital to Darbhanga" gives **HBCH Muzaffarpur (90), Mahavir Cancer Sansthan (70), Paras HMRI Patna (50)** plus national referrals. Nearer TMC unit first, Patna as alternatives rather than the default.

## Proximity ranking: assessed and deliberately not built (P1-12)

Three data-backed reasons:

- **No geographic primitive.** No `district`, no coordinates. `region` is unusable — it splits *within* Bihar (HBCH Muzaffarpur is `"East India"`, AIIMS Patna is `"East"`).
- **Catchment prose is too sparse, and misses the reported city.** Only 2 of 12 Bihar records name a catchment, and **Darbhanga is in neither** — so a catchment-driven map would not have fixed the query in this issue. The only Darbhanga signal in the file is HBCH's `logistics.nearest_airport: "Darbhanga Airport (~50km)"`.
- **It would be clinically wrong.** Distance-ranking eastern Bihar promotes Healing Touch Bhagalpur (tier C, score 50) over HBCH (tier A, 90, TMC) — and Healing Touch's own record says *"No radiation or medical oncology — refer to Patna or Muzaffarpur for those needs."* Exactly the failure this issue forbids.

Recommended order: capture `district` and structured `catchment_districts` **through the Navigator review/approve pipeline**, then use proximity only as a tiebreak *within* a score band, never as the primary key, and never synthesise distances from the `nearest_railway`/`nearest_airport` strings. The new fallback logging makes it possible to count which districts actually generate fallbacks before anyone invests.

## Tests

`hospital-directory.service.spec.ts` gives **50/50** (21 new). Base proof: with the two source files reverted to `e63a99a` and the new spec kept, **16 of 21 fail**, including all eight "non-empty set / HBCH ranked first" assertions for Darbhanga, Samastipur, Purnia and Motihari. Full API suite **914/914**, `tsc --noEmit` clean.

## Eval cases — and two honesty caveats

`HN-DARBHANGA-01` (English) and `HN-DARBHANGA-02` (Hinglish) added to `eval/cases/tier1/hospital_navigation.yaml`, matching the neighbouring cases' schema exactly and reusing the existing `CARE_NAVIGATION_PROVIDER_CHOICE` intent and `suchi.hospital_navigation.v1` rubric. **No new intent, rubric or clinical pass criteria invented**, so no SCCF gate was hit. Manifest regenerated with the repo's own tool: 601 to **603**, `check` green; the pinned counters in `eval/scripts/case-schema.test.ts` and `docs/CAPABILITY_LEDGER.md` updated to match. Eval framework suite **70/70**.

1. **The gold-lane runner scores from the *rubric*, not the `expectations` block** — so `must_mention_any`, `must_contain_any` and `hospital_count_min` in this file are documentary, not executed. The HBCH-specific assertion is enforced by the **unit test**, not by the eval case. A header comment now records which keys run and which do not, so the next person does not repeat #89. Making the eval assert HBCH would need a new rubric check, which changes scoring for 5 existing cases — not done here.
2. **`npm run eval:tier1` only runs `retrieval_quality.yaml`**, so `hospital_navigation.yaml` — all 7 cases, including the 5 that pre-date this PR — is **not in the nightly**. Adding it is a lane decision, left alone.

## Hindi path — split result (P1-13)

- **Hinglish reaches the same `searchHospitals` call and is fixed here.** `detectLocation("Darbhanga ke paas cancer hospital")` gives `{city:"Darbhanga", state:"Bihar"}`, identical hospital list to the English query. Acceptance criterion met for Latin-script Hindi.
- **Devanagari never reaches it at all.** `execution-planner.service.ts` gates the lookup on a pattern using `\b` beside a Devanagari term, and JS `\b` never matches next to Devanagari — so it cannot fire. Confirmed via the planner: even when handed `category: "NAVIGATION"`, `structuredHospitalResults` is `null` for the Devanagari query. Second independent gap: `INDIAN_CITIES` has Latin aliases only, so the Devanagari city name returns `null` regardless.

**Not fixed here, deliberately.** This is the same `\b`-on-Devanagari root cause already documented as QA0904-1, it sits in detection patterns adjacent to the safety keyword lists, and the Devanagari term list needs SCCF review (AGENTS.md section 1.3). It should be fixed once, consistently, not patched per call site.

## Notes

- **`hospitals.json` untouched.** Also a correction to the record: there are not two copies — `apps/api/data/hospitals.json` is a **symlink** to `apps/landing/src/content/hospitals.json`. One file, 83 entries.
- Noted but not changed: `query-expander.service.ts:207` has a Patna-only `'hospital in patna'` expansion and no Darbhanga entry. That is RAG retrieval, out of scope — and the structured block now populating for all Bihar cities largely defuses it, since the block instructs the model not to mention unlisted hospitals.
- **Merge-order note:** PR #59 also moves the pinned case counter (to 604/9/595 for its own 3-case fixture). Whichever of the two merges second needs the pin at **606/9/597**. Both edit the same `expect` block, so git will conflict rather than fail silently.


gautamgauri:
Holding this PR — I confirmed the first review blocker, and it is real and introduced by this change.

## "Distant fallback hospitals can be labelled nearby" — confirmed

`chat.service.ts:3158` builds the prompt block as:

```
--- Regional / Nearby Centres ---
```

Before this PR, an unmatched city produced zero rows and `buildHospitalContextBlock()` returned `""`, so no block was emitted at all. With the fallback chain, the block now populates — and at the **adjacent-state rung** it will list hospitals 300–400km away under a heading that says *Nearby*. For a Darbhanga query that widens into West Bengal, Siliguri or Kolkata would be presented to the model as nearby centres.

Within Bihar the label is defensible (Patna from Darbhanga is ~150km, and "regional" is fair). It is the cross-state rung that makes the heading false. So the fallback fixed the empty-set bug and introduced a truthfulness bug in its place, on a patient-facing navigation path where the whole value is "where should I go from where I am".

## Fix I intend

Label the block by the fallback stage that actually produced it, rather than always "Regional / Nearby":

- exact city → `Centres in <city>`
- same state → `Centres in <state>`
- adjacent state → `Centres in neighbouring states — may involve significant travel`
- unfiltered pool → `Other centres — travel distance not established`

`resolveGeographicCandidates()` already knows the stage (it logs it as `hospital_search_geographic_fallback`), so this is threading a value that exists, not inventing geography. That keeps the candidate-set fix while making the heading tell the truth about how far the widening went.

## Second blocker — "regression cases do not enforce Muzaffarpur"

Partly answered in the PR body, but worth restating plainly: the **unit tests do** enforce it (`hospital-directory.service.spec.ts` asserts a non-empty set with HBCH ranked first for Darbhanga, Samastipur, Purnia and Motihari, and 16 of 21 fail on base). The **eval cases do not** — the gold-lane runner scores from the rubric, so `must_mention_any` / `hospital_count_min` in `hospital_navigation.yaml` are documentary. Making the eval enforce HBCH needs a new rubric check, which changes scoring for 5 existing cases.

So the enforcement exists, just not where a reader of the eval file would look. If the requirement is that the *eval* enforces it, that is a rubric change and should be its own PR with SCCF sight of the scoring change.

Will push the heading fix here rather than opening another PR, since it is a direct consequence of this change.

gautamgauri (commented):
Still holding this until the fallback stage is propagated into the patient-facing context label. The candidate widening fixes the Darbhanga empty-set defect, but an adjacent-state or unfiltered fallback must not be presented as simply “Regional / Nearby Centres.” The heading should disclose the rung actually used, as already proposed in the review thread.

On the second finding: I’m fine keeping HBCH enforcement in the unit layer for this PR if the eval limitation is stated accurately. The new eval cases should not be described as guarding the Muzaffarpur outcome until an executable rubric/deterministic check exists. That can be a separate eval PR.

gautamgauri:
## Rebased onto current `main` (`5a2d4f7`) — conflict resolution notes

Force-pushed with lease. **Not merged.** Three files conflicted, all from **#107** ("make the case manifest count what runs, not what parses") landing on `main` today.

### 1. `eval/scripts/case-schema.test.ts` — resolved to main's shape

`main`'s pin is now `{total, runnable, unrunnable, blockers}` via `summariseExecutability`, not this PR's older `{total, unrunnable, runnable: total - unrunnable}`. Kept main's shape and added this PR's 2 runnable gold-lane cases:

```
total: 603, runnable: 574, unrunnable: 29,
blockers: { "orphan-schema": 9, "missing-rubric": 20 }
```

`blockers` is unchanged because `HN-DARBHANGA-01/02` reuse `CARE_NAVIGATION_PROVIDER_CHOICE`, which `rubrics.v1.json` defines. **This was not assumed** — `npm --prefix eval run cases:update` independently reports `574 executable of 603 cases present in 25 files — 29 blocked (20 missing-rubric, 9 orphan-schema)`, matching exactly.

### 2. `eval/cases/case-manifest.json` — regenerated, not hand-merged

Took main's `version: 2` header and re-ran the repo's own generator rather than resolving by hand, so the new `executableCases`/`blockers`/per-file fields are tool-produced.

### 3. `docs/RELIABILITY_BACKLOG.md` — renumbered to P1-12 / P1-13

`main` now carries a **P1-11** from the #94 follow-up, which this PR also claimed. Kept main's P1-11 and moved this PR's two entries to **P1-12** (hospital proximity) and **P1-13** (Devanagari) — matching what this PR's own follow-up commit `2f457c1` already intended. That commit therefore became a no-op against the resolved tree and was dropped from the rebase; its content is fully present. Also updated the decision-list reference from "quote executable coverage as 572, not 601" to "574, not 603".

The **merge-order note in the PR description is now stale** — it says the second of #59/#99 to merge needs the pin at `606/9/597`. Under #107's shape that becomes **total 606 / runnable 577 / unrunnable 29**, not `606/9/597`.

### Verification after rebase

| suite | result |
|---|---|
| `cd apps/api && npx jest` | **52 suites / 996 tests pass** |
| `cd eval && npm test` | **8 suites / 94 tests pass** |
| `cd eval && npm run typecheck` | clean |
| `npm --prefix eval run cases:update` | 574/603, 29 blocked — consistent with the pin |
| CI `API unit tests` | pass |
| CI `Build + config parity` | pass |

### Pre-existing inconsistency, noted not fixed

`main` labels **two different items P1-11**: the section heading is the #94 follow-up, while numbered decision item 12 calls the missing-rubric issue "P1-11". Both #94 and #89 claimed the id. I did not renumber main's entries — that is outside this PR's scope, and touching it here would add noise to a rebase. Worth a one-line cleanup separately.

gautamgauri:
## Tier1 verdict for this PR: **fail (6 of 21, 91% avg) — pre-existing, not caused by this change**

Adding this so the CI state on this PR is not left ambiguous. Required checks (`API unit tests`, `Build + config parity`) are both **green**; the rebase and conflict-resolution notes are in my [earlier comment](https://github.com/gautamgauri/suchi-cancer-bot/pull/99#issuecomment-5607436227).

Evidence it is not this PR (full analysis in [my comment on #98](https://github.com/gautamgauri/suchi-cancer-bot/pull/98#issuecomment-5607626168)):

- `.github/workflows/eval-tier1.yml:81` points `EVAL_API_BASE_URL` at the **live deployed Cloud Run service** — on a `pull_request` event the branch is never built or served, so the check measures **production**, not this code.
- **#58, which changes only line endings, scores the identical 6/21** against the same endpoint.
- The failing cases are all `RQ-*` **retrieval** cases (`RQ-PANCREAS-01`, `RQ-BREAST-02`, `RQ-PROSTATE-02`, `RQ-LIVER-01`, +2) failing **LLM-judge** sub-checks. This PR touches hospital-directory candidate selection and does not participate in those paths at all.
- Some of the 6 are not quality failures: on #98's run `RQ-PROSTATE-02` had all six judge checks return Vertex AI **`429 RESOURCE_EXHAUSTED`**, which the runner scores as *failed* rather than *errored*, dragging it to 36.4%.

**Worth noting for this PR in particular:** even if Tier1 *did* run against the branch, it would not exercise this change. As the PR body already records, `npm run eval:tier1` runs only `retrieval_quality.yaml`, so `hospital_navigation.yaml` — including the new `HN-DARBHANGA-01/02` and the 5 pre-existing cases — is not in that lane. The real verification for this PR is the **unit suite**: `hospital-directory.service.spec.ts` at 50/50, with the recorded base proof that 16 of 21 new assertions fail when the two source files are reverted to `e63a99a`.

gautamgauri (commented):
**Still HOLD on the current head.** The candidate-set fix is useful and the Darbhanga unit coverage is strong, but `searchHospitals()` still discards the `GeographicStage` returned by `resolveGeographicCandidates()`. That means `buildHospitalContextBlock()` cannot distinguish exact-city/state results from adjacent-state or fully unfiltered fallbacks, so distant facilities can still be presented under the existing “Regional / Nearby Centres” framing.

Please propagate the fallback stage into the patient-facing context label (or suppress the unfiltered regional block and rely on national referrals). I’m fine keeping the HBCH/Muzaffarpur assertion in the unit layer for this PR as long as the eval cases are described accurately as documentary rather than executable enforcement.

gautamgauri:
Review status: **HOLD — truthfulness blocker remains, and #124 should land first.**

The geographic fallback logic now returns `{ results, stage }`, but `searchHospitals()` immediately discards the stage:

```ts
results = this.resolveGeographicCandidates(results, params).results;
```

So the downstream prompt still cannot distinguish exact-city/same-state results from adjacent-state or fully unfiltered fallbacks. That means distant hospitals can still be presented under the existing `Regional / Nearby Centres` framing even though the code knows the search widened beyond anything we have established as nearby.

Please thread `GeographicStage` through to the context-builder and label the block truthfully, e.g. city/state vs `neighbouring states — may involve significant travel` vs `other centres — travel distance not established`.

Also: merge/deploy #124 first. Until the API image actually contains the 83-hospital directory, this PR cannot improve the production answer path at all.

gautamgauri:
Fresh review against current `main` (`6ca440d`, after #131/#132/#133): **still HOLD, but the old #124 prerequisite is now satisfied.**

The hospital directory is already shipped in the API image, so that part of the previous review is no longer a blocker. The remaining functional blocker is unchanged and introduced by this PR's fallback design:

- `resolveGeographicCandidates()` knows whether results came from exact city / same state / adjacent state / unfiltered pool;
- `searchHospitals()` still discards that `stage` and passes only `.results` downstream;
- the prompt therefore cannot tell a same-state alternative from a cross-state/unbounded fallback and can still label distant centres as `Regional / Nearby Centres`.

Please thread the stage into the context builder and render truthful headings, e.g. city/state vs `neighbouring states — may involve significant travel` vs `other centres — travel distance not established`.

Also refresh/rebase this branch onto current `main` before merge; it is currently non-mergeable and predates the recent retrieval/FTS work. No need to revisit the candidate ranking itself — the blocker remains the patient-facing distance/truthfulness framing.

gautamgauri:
## Rebased onto `main` @ `6ca440d`, and the stage blocker is addressed

`mergeable` is `true` again. Two commits on top of `6ca440d`.

### Rebase

Single conflict, in `apps/api/src/modules/chat/hospital-directory.service.ts`, against **#124**'s loader refactor — `main` now imports `readHospitalDirectoryFile` / `recordHospitalDirectoryStatus` from `common/hospital-directory-file` where this branch's base still read the file inline. **Both kept**: #124's loader and status recording are intact in `onModuleInit()`, and this PR's `resolveStateForCity` import sits alongside them. `git range-diff` against the pre-rebase head shows the only differences are those import-context shifts and two hunk headers in `docs/RELIABILITY_BACKLOG.md`; no logic was dropped. The #124 prerequisite you raised on 2026-09-12 is satisfied by that same commit already being on `main`.

### The HOLD item: the stage is no longer discarded

> `searchHospitals()` still discards that `stage` and passes only `.results` downstream

Fixed. `searchHospitalsWithGeography()` now returns `{ results, geography }`, where `geography` is `{ stage, requestedCity, resolvedState }`, and that value is threaded `ExecutionPlan` → `PlanExecutionResult` → `ChatService` → `buildHospitalContextBlock()`. `searchHospitals()` stays as a thin delegate so callers that only want rows are unaffected. `GeographicStage` is no longer a dead export.

`--- Regional / Nearby Centres ---` is gone. The heading is derived from the rung that actually produced the set:

| stage | heading |
|---|---|
| `city` | `Centres in <city>` |
| `state` | `Centres in <state>` |
| `adjacent_state` | `Centres in neighbouring states — may involve significant travel` |
| `unfiltered` | `Major cancer centres (none found near <city>) — travel distance not established` |
| `none` (no location given) | `Major cancer centres` |

On the `adjacent_state` and `unfiltered` rungs the block additionally carries an explicit instruction — no cancer centre in the directory serves `<city>` directly, do not describe these as "nearby", and do not state or estimate a travel time or distance. The heading alone is easy for a generator to paraphrase away; the instruction is not.

**No heading claims a distance or a travel time.** The directory holds no coordinates, so any such claim would be invented. That is #103's work, and it is next.

### Which review items this addresses

- ✅ **P1, Codex `hospital-directory.service.ts:297`** — "preserve the fallback stage before labeling results nearby". The Jaipur/Rajasthan case in that comment is covered directly: it lands on `unfiltered` and is now headed `Major cancer centres (none found near Jaipur) — travel distance not established`, with the caveat block.
- ✅ **Your 2026-09-09, 2026-09-11, 2026-09-12 and 2026-09-12 (21:32) HOLDs** — all the same item, the stage not reaching the patient-facing label.
- ✅ **Non-mergeable / predates the retrieval+FTS work** — rebased onto `6ca440d`.
- ⬜ **P2, Codex `hospital_navigation.yaml:84`** — "enforce the Darbhanga hospital expectation". Unchanged and still accurate as you framed it: enforcement lives in the unit layer, the eval cases are documentary. Making the eval assert HBCH needs a new rubric check that rescores 5 existing cases, which is a separate PR with SCCF sight of the scoring change. The file's header comment already records which keys execute and which do not.

The never-empty guarantee from #95 is untouched — the `unfiltered` rung still returns the full regional pool rather than `[]`, it is just labelled honestly now.

### Tests

- `hospital-directory.service.spec.ts`: **10 new** stage-propagation cases — one per rung, state recovered from the city table, `searchHospitals()` proven to return exactly the rows of the geography variant, the stage proven to describe the pool the caller asked for rather than the post-cancer-type-filter set, and the real shipped directory reporting Darbhanga / Samastipur / Purnia / Motihari as **same-state widenings** rather than exact-city matches (with Muzaffarpur as the `city`-stage control).
- New `chat.service.hospital-context.spec.ts`: **12 cases** covering every stage's heading, the travel caveat appearing on exactly the two rungs that need it, the national-referral block staying separate, and a loop asserting the old proximity-asserting heading cannot be produced on **any** stage.
- `npx jest --testPathPatterns="hospital|location|chat.service"` → **8 suites / 127 tests pass**.
- Full API suite → **1090 of 1096 pass**. The 6 non-passing are `whatsapp.concurrent-inbound.spec.ts` (5 failed, 1 skipped) timing out at 5000ms under full-suite load; that suite passes **6/6 in isolation** and shares no code with this change.
- `npm run build` (`nest build`) clean.

No hospital data changed, no clinical or KB content touched.

gautamgauri:
Updated review after the rebase: **HOLD — the original geography-label blocker is fixed, but the fix now crosses the repo's prompt-review boundary.** The current head correctly threads the fallback stage and replaces the false `Regional / Nearby Centres` heading with truthful city/state/adjacent/unfiltered labels. That part looks good. However, `chat.service.ts` also adds a new LLM instruction (`Do NOT describe them as nearby... do NOT state or estimate...`). AGENTS.md explicitly requires prompt changes under `chat/` to go through separate SCCF human/medical review. Please either keep this PR structural only (stage-aware headings/data, no new prompt instruction) or split the instruction into a clearly labelled SCCF-review PR. The HBCH/Muzaffarpur unit coverage is sufficient for this PR; the documentary eval limitation can remain a separate eval issue.

PR #100, opened 2026-09-08
fix(whatsapp): serialise one contact inbound burst and close the resolveSession race (#94 follow-up)

gautamgauri:
Follow-up to #97, split out per review so the safety-relevant template correction could ship on its own.

#97 is now merged to `main` (squash commit 913f26c), and this PR has been rebased onto `main` — the diff is exactly the two WhatsApp files (queue work only).

## Two defects, both from the 2026-09-06 burst

- **Unbounded concurrency per contact.** Four webhook POSTs each fired their own fire-and-forget turn, so one contact ran four full chat turns at once. That amplified latency to 25-33s, which is what put a turn within reach of the 30s pre-RAG budget guard that #97 fixes.
- **`resolveSession` is read-then-write**, so one new contact burst minted **four sessions**. Not reported in #94 - found while writing the reproduction.

## Change

One promise chain per `waId`. A contact messages process one at a time, different contacts still run in parallel, batch order is preserved, and a failure cannot wedge the queue. Instance-local by design; the wamid ledger remains the at-most-once guarantee across instances.

## Tests

`whatsapp.concurrent-inbound.spec.ts` - the rapid-sequential burst test. On base, 3 of 6 assertions fail (four concurrent turns for one contact; four sessions for one new contact). Note the text-pairing assertion **passes on base**: that is the negative result that disproved the original hypothesis in #94.

Full API suite green on the rebased head: **48 suites, 905 tests passed**, including the 6 new concurrent-inbound tests.

## Explicitly out of scope

Does **not** fix the ~25-33s base turn latency. Serialisation removes the amplification, not the latency. A per-phase budget is still needed before wider WhatsApp rollout - tracked as P1-11.

Also unaddressed here, per review: whether an urgent follow-up message should be able to jump the queue ahead of a slow in-flight reply. Ordering guarantees versus urgency pre-emption is a product/safety tradeoff, not a plumbing choice, and it wants its own decision.

Co-authored with Claude Code.

### DIFF
--- a/apps/api/src/modules/whatsapp/whatsapp.concurrent-inbound.spec.ts
+++ b/apps/api/src/modules/whatsapp/whatsapp.concurrent-inbound.spec.ts
@@ -1,0 +1,237 @@
+import { WhatsAppController } from "./whatsapp.controller";
+import { WhatsAppService } from "./whatsapp.service";
+import { MetaWebhookBody } from "./whatsapp.types";
+
+/**
+ * Back-to-back inbound messages (issue #94).
+ *
+ * On 2026-09-06 a user sent "Hii", "Hi", "Hey", "Hello" within seconds and one
+ * reply was the S2 urgent-escalation template. The leading hypothesis was that
+ * concurrent `processClaimed` calls crossed their wires and fed one claim's text
+ * to another claim's pipeline. These tests check that directly (they pass on the
+ * unfixed code — the texts never cross) and pin the concurrency behaviour that
+ * *was* wrong: four full chat turns for one contact ran simultaneously, which
+ * raced session minting and stretched a turn past ChatService's 30s pre-RAG
+ * budget.
+ */
+
+/** Ledger stand-in: PK collision → Prisma P2002, as the real de-dup relies on. */
+function makeLedgerMock() {
+  const rows = new Map<string, any>();
+  return {
+    rows,
+    create: jest.fn(async ({ data }: any) => {
+      if (rows.has(data.id)) {
+        const err: any = new Error("Unique constraint failed on the fields: (`id`)");
+        err.code = "P2002";
+        throw err;
+      }
+      rows.set(data.id, { ...data });
+      return data;
+    }),
+    findUnique: jest.fn(async ({ where }: any) => rows.get(where.id) ?? null),
+    update: jest.fn(async ({ where, data }: any) => {
+      const row = { ...(rows.get(where.id) ?? { id: where.id }), ...data };
+      rows.set(where.id, row);
+      return row;
+    }),
+  };
+}
+
+function textWebhook(wamid: string, from: string, body: string): MetaWebhookBody {
+  return {
+    object: "whatsapp_business_account",
+    entry: [
+      {
+        changes: [
+          {
+            field: "messages",
+            value: {
+              messaging_product: "whatsapp",
+              metadata: { phone_number_id: "PNID" },
+              messages: [{ id: wamid, from, type: "text", text: { body } }],
+            },
+          },
+        ],
+      },
+    ],
+  } as MetaWebhookBody;
+}
+
+const REQ = { rawBody: Buffer.from("{}") } as any;
+const CONTACT = "919999900001";
+const BURST = [
+  { wamid: "wamid.A", text: "Hii" },
+  { wamid: "wamid.B", text: "Hi" },
+  { wamid: "wamid.C", text: "Hey" },
+  { wamid: "wamid.D", text: "Hello" },
+];
+
+describe("WhatsApp rapid-sequential inbound (issue #94)", () => {
+  let prisma: any;
+  let chat: any;
+  let sessions: any;
+  let svc: WhatsAppService;
+  let ctrl: WhatsAppController;
+  let sent: Array<{ to: string; text: string }>;
+  /** Chat turns currently executing, per contact. */
+  let inFlight: Map<string, number>;
+  let maxInFlight: Map<string, number>;
+  let sessionSeq: number;
+
+  beforeEach(() => {
+    process.env.META_WABA_PHONE_NUMBER_ID = "PNID";
+    process.env.META_WABA_TOKEN = "tok";
+    process.env.META_APP_SECRET = "secret";
+    process.env.WHATSAPP_VERIFY_TOKEN = "verify";
+    process.env.WHATSAPP_SESSION_TTL_HOURS = "24";
+
+    sent = [];
+    inFlight = new Map();
+    maxInFlight = new Map();
+    sessionSeq = 0;
+
+    // A contact row that only becomes visible once it has been written, so a
+    // racing `resolveSession` genuinely mints a second session.
+    const contacts = new Map<string, any>();
+    prisma = {
+      whatsAppContact: {
+        findUnique: jest.fn(async ({ where }: any) => contacts.get(where.waId) ?? null),
+        update: jest.fn(async ({ where, data }: any) => {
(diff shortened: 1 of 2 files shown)


gautamgauri:
On hold pending a decision on urgency behaviour — converted to draft so it is not reviewed as ready.

## The recommendation on the table

Preserve ordering for ordinary messages, but assess urgent follow-ups promptly using the existing safety rules.

I think that is the right shape. Two constraints on it that are worth settling before anyone implements, because both are easy to get wrong:

**1. "Promptly" must mean re-order, not run in parallel.** Unbounded concurrency per contact is what amplified turn latency to 25–33s, and that latency is what put a turn within reach of the 30s pre-RAG budget guard — the branch that produced the false escalation in #94 in the first place. If an urgent follow-up pre-empts by starting a *second concurrent turn*, we reintroduce the amplifier and can land back on the same guard. Safe version: keep one in-flight turn per contact and let an urgent message jump to the **front of that contact queue**, or cancel the in-flight non-urgent turn. Either preserves the invariant this PR establishes; a parallel fast path does not.

**2. The urgency signal is currently uneven across languages.** Detection would lean on `emergency-fast-path.ts` / `safety.rules.ts` / `hasUrgencyIndicators()`. Those are known to miss Hinglish (#81 P0: post-chemo `bleeding bahut zyada` + `chakkar` classified `normal`) and cannot fire at all on Devanagari, where `\b` never matches beside the script (P1-13, same root cause as QA0904-1). So queue pre-emption keyed on those rules would work in English and silently not work for a Hindi-speaking patient — an uneven safety guarantee, on the channel whose users are most likely to write in Hindi. That is an argument for sequencing the Hinglish/Devanagari coverage work alongside this, not for blocking it.

Worth noting the scope: re-ordering a queue is plumbing and needs no new escalation vocabulary or template, so it does not by itself require SCCF sign-off. But it changes *when* an escalation reaches a patient, which is safety-relevant behaviour, so the decision should be recorded rather than absorbed into an implementation PR.

The serialisation in this PR is unaffected by either point and remains correct as the base invariant.

gautamgauri:
Correction accepted, and I traced what cancellation would actually require. It is more constrained than my previous comment implied.

## Front-of-queue does not deliver prompt escalation

Right: jumping the queue still leaves an urgent message behind the in-flight turn, which on this channel is the 25–33s turn. So it raises priority and guarantees nothing about latency. I should not have offered it as an alternative to cancellation — they are different guarantees.

## True cancellation is a multi-layer change, not a flag

Traced on `main` today:

| Layer | State |
|---|---|
| `ChatService.handle(dto, signal?)` | accepts an `AbortSignal` |
| `chat.controller.ts:25` (web) | passes `abortController.signal` |
| **`whatsapp.service.ts:202,210`** | **passes no signal at all** |
| `voice.service.ts:196` | passes no signal |
| **`llm.service.ts`** | **no method accepts an `AbortSignal`** — its `AbortController`s are local timeout controllers (`:1242`, and the `signal: controller.signal` sites at `:1115`/`:1353` are on the legacy OpenAI/DeepSeek paths, not the live Gemini one) |
| `rag.service.ts` | never sees a signal |
| abort checkpoints inside `handle` | **2** (`:125`, `:912`) against **16** `prisma.message.create` sites |

Two consequences, both of which are the failure mode called out in review:

1. **Abandoning the promise does not stop the work.** The caller signal cannot reach the LLM request, because `llm.service` has no parameter to receive it. The turn keeps running to completion regardless.
2. **A cancelled turn can still reply.** With 2 abort checks and 16 reply-write paths, an abandoned turn can persist an assistant row and send an ordinary reply after the urgent one — the late-reply hazard, concretely.

So the honest scope is: thread a signal from the WhatsApp path, give `llm.service` (and the RAG entry points) a signal parameter, and gate every reply-write on abort — before any urgency pre-emption is trustworthy.

## Recorded limitation

**This PR does not provide prompt urgent escalation, and no configuration of it does.** It bounds concurrency, which is what #94 required; the time an urgent follow-up waits is still the remaining in-flight turn.

## Test requirement before any urgency claim

When urgency pre-emption is built, it must be tested explicitly in **English, romanized Hinglish, and Devanagari**, and **no equivalent-urgency-handling claim should be made until all three pass**. On current `main` two of the three are expected to fail:

- Hinglish — #81 P0, post-chemo `bleeding bahut zyada` + `chakkar` classifies `normal`
- Devanagari — P1-13 / QA0904-1, `\b` cannot match beside the script, so the rules cannot fire at all

A pre-emption keyed on those rules would therefore work in English and silently not work in Hindi. Tests across all three scripts are what stop that shipping as "urgent messages are handled promptly".

gautamgauri:
Retargeted to `main` now that #97 has merged.

⚠️ **This PR needs a rebase before review.** #97 was squash-merged, so `main` carries the squashed `913f26c` while this branch still has the original unsquashed timeout commit. Git therefore sees them as unrelated histories and the diff has grown from 2 files to **5** — it is re-proposing the timeout fix that is already on `main`.

Not rebasing yet, deliberately: this PR is on hold pending the urgency decision, and that decision will change the code here (signal threading, abort-gated reply writes). Rebasing now would just have to be redone. Flagging it so nobody reads the 5-file diff as this PR scope, and so it is not merged in that state.

The queue change itself — `whatsapp.service.ts` plus `whatsapp.concurrent-inbound.spec.ts` — is unchanged and remains the 2 files described above.

gautamgauri (commented):
Keep this in draft. The urgency/pre-emption decision is still worth settling, but there are also two plumbing blockers independent of that decision:

- acknowledged messages queued only in an in-memory promise chain are not recoverable after instance termination; with 25–33s turns, serialisation increases that loss window;
- separate webhook POSTs can enter the queue in ledger-insert completion order rather than actual inbound order, so two patient turns can be inverted.

Please solve durability-before-ACK and add a deliberately reversed ledger-latency test before treating the serial queue as production-safe. Any urgency pre-emption should still preserve one in-flight turn per contact rather than reintroducing parallel turns.

gautamgauri:
**Product decision (Gautam, 2026-09-09): keep strict sequential ordering for now — no urgency pre-emption.**

An urgent message does not jump the queue ahead of a slow in-flight turn. Rationale: out-of-order replies risk re-introducing the interleaved-session-state class of bug behind #94; the safety fast path is preserved for messages once their turn runs. Deferred (not rejected): urgency pre-emption via a fast safety pre-check on inbound, to be revisited with SCCF if real-world latency of red-flag escalations proves material. Complementary: #81 keyword coverage must land for detection to fire at all on Hinglish red flags.

This settles the urgency question. Remaining before merge: the two Codex P1 plumbing findings — (1) queued-turn durability (acked-but-unprocessed messages live only in the in-memory chain; needs a privacy-aware persistence design), (2) enqueue in webhook-arrival order, not claim-completion order.

gautamgauri:
**Decision (Gautam, 2026-09-09): log as accepted WhatsApp-channel constraints — no code change for now.**

Both Codex P1 findings are channel-adapter properties (Meta ack semantics + concurrent POST bursts), not core-pipeline defects. Accepted for the current live channel:

1. **Queued-turn durability window** — acked-but-unprocessed turns exist only in the instance-local promise chain; an instance death between ack and processing can lose a queued patient message (Meta will not redeliver an acked event). Accepted at current scale; tracked.
2. **Claim-order race** — across separate rapid POSTs for one contact, enqueue order follows ledger-insert completion, not webhook arrival. Accepted; the serialisation itself remains strictly better than the pre-fix state.

Revisit trigger: before wider WhatsApp rollout or if per-contact burst volume grows. Sequential ordering (no urgency pre-emption) stands. PR #100 is unblocked for merge as-is.

Issue #68, opened 2026-08-22
Citation strip is fail-open: a truncated marker reaches the reader


gautamgauri:
`chat.controller.ts` strips citation markers from display text on purpose — the auditors-not-readers decision recorded in #54. All three patterns require a **closing bracket**:

```js
const CITATION_MARKER_PATTERN     = /\[citation:[^\]]+\]/g;
const NUMBERED_REF_PATTERN        = /\s*\[\d{1,3}\]/g;
const RAW_SOURCES_SECTION_PATTERN = /\n\n\*\*Sources:\*\*\s*(\[citation:[^\]]+\]\s*)+/g;
```

When a generation stops *inside* a marker there is no closing bracket, nothing matches, and the raw knowledge-base identifier survives the strip and reaches the reader. Observed on the live service:

```
...might be effective [citation:kb_en_nci_types_breast_diagnosis_breast_cancer_biomarker_tests_v1:kb_
```

`utils/citationParser.ts` has the same gap on the client side — `CITATION_PATTERN` also requires `]`, so the unterminated marker falls through `splitTextWithCitations` into the plain-text branch and is rendered verbatim.

### Why it's worth fixing even though it's rare
Seen once in four runs, so it is not frequent. But the failure is on the trust surface: a reader who is told "this is educational information, consult your doctor" and then sees `kb_en_nci_types_breast_diagnosis_...` learns something about the system that no user should have to see. It fails open rather than closed.

### Suggested direction
One added pattern handles it in both places — an unterminated marker running to end-of-text, e.g. `/\[citation:[^\]]*$/`. Stripping is the safe direction: a citation the reader was never meant to see costs nothing when removed.

---
*Found by driving the live service at `suchi-web` through a real browser on 2026-08-22, across four runs of the nine `phase2_journeys.yaml` cases. Answers are not deterministic — a defect seen once is real; a defect not seen once is not absent.*

PR #106, opened 2026-09-09
fix(voice): stop citation markers and raw markdown reaching TTS (#87)

### LINKED ISSUE #87: Voice/TTS can read citation markers and raw markdown aloud — two independent paths
## Summary

Two separate defects, found independently by two agents working different issues on 2026-09-06, both ending in the same place: **raw KB identifiers or markdown reaching text-to-speech.** Neither is fixed; both were deliberately left out of scope of PRs #82 and #83 to keep those scoped.

A reader who is told "this is educational information, consult your doctor" and then *hears* `kb_en_nci_types_breast_diagnosis_breast_cancer_biomarker_tests_v1` learns something about the system no user should have to. Same trust failure as #68, on a surface where it cannot be skimmed past.

## Path 1 — the voice stripper has the same fail-open gap as #68

#68 was about the display strip requiring a closing `]`, so a generation stopping mid-marker leaked the raw id. PR #82 fixes that for the **display** surface (`chat.controller.ts` → new `display-text-cleaner.ts`) and the **web client** (`citationParser.ts`).

The identical gap remains in the voice path, all still requiring `]`:

- `apps/api/src/modules/voice/voice-output-stripper.ts:44`
- `apps/api/src/modules/voice/voice.service.ts:235-236`
- `apps/api/src/modules/voice/voice-response-condenser.service.ts:17`

And in two copilot services:

- `copilot/services/patch-executor.service.ts`
- `copilot/services/eval-bridge.service.ts`

PR #82's fix is directly portable — it restricts unterminated-marker content to the real doc/chunk-id charset so the strip halts at the first space and cannot swallow following prose. PR #82 also closed a latent data-loss bug worth checking for here: patterns that did not exclude `[` matched `[citation:doc1 …prose… [citation:doc2:chunk2]` as a single marker and **silently deleted the sentence in between**.

## Path 2 — the structured-template branch skips `stripForVoice` entirely

Separate and arguably worse, because no amount of hardening the stripper fixes it: the stripper is never called.

`chat.service.ts:1186` returns early on the structured-template branch, before the voice stripping step. So a request with `channel: "voice"` that routes to a template sends **raw markdown and intact `[citation:…]` markers** straight to TTS — not a truncated fragment, the complete markers.

Found by the agent on #67/#69 while tracing why `removeDuplicateBullets` never runs on that path (`deduplicateResponse` sits at `:2207`, well after the `:1186` return). Same early-return, two different pieces of post-processing skipped.

## Why this has not been noticed

Voice has no equivalent of the browser probes that found #66–#71. `eval/cases/voice/` exists and `voice-transcript-eval.ts` has a `must_not_mention` check — the only place in the eval framework where `must_not` is actually enforced — but nothing asserts the absence of citation markers or markdown in TTS-bound text.

## Suggested direction

1. Port PR #82's `display-text-cleaner` logic to the voice stripper, or better, have both call one shared cleaner so the two surfaces cannot drift apart again.
2. Make the `chat.service.ts:1186` early return honour `channel: "voice"` — either strip before returning, or move the branch so post-processing is unconditional. Worth checking what *else* that early return skips beyond voice stripping and bullet dedup.
3. Add a voice-path assertion that TTS-bound text contains no `[citation:`, no `**`, and no markdown headings. A `must_not_mention` entry would work with the existing machinery.

## Not in scope here

The clinical content of voice replies. This is purely about post-processing leaking machine identifiers and markup into speech — no safety or clinical copy is involved.

## References

- #68 / PR #82 — the display-surface version of Path 1
- #67, #69 / PR #83 — where Path 2 was found
- FR-WA-011 records the same policy for WhatsApp: citation markers are never shown to end users

gautamgauri:
Closes #87.

Two independent defects, both ending with machine identifiers or markup reaching text-to-speech. Both are closed here. **No clinical content, prompt wording, or safety-module code is touched** — this is post-processing only.

## Path 1 — fail-open marker patterns on the voice surface

Every citation strip in the voice path required a closing `]`, so a generation that stopped mid-marker spoke the raw KB id aloud:

```
Biomarker tests guide treatment [citation:kb_en_nci_types_breast_diagnosis_breast_cancer_biomarker_tests_v1:kb_
```

And `[^\]]*` did not exclude `[`, so an unterminated marker matched forward into the next complete one and **silently deleted the prose in between**:

```
in:  [citation:doc1 Radiotherapy is given daily. [citation:doc2:chunk2]
out: ""            <- the whole sentence, gone
```

These are the same two defects PR #82 fixed for the display surface (#68). Rather than porting the patterns a second time and letting the two copies drift again, the hardened patterns move into `apps/api/src/common/text-cleaning.ts` and **both** surfaces call it. `display-text-cleaner.ts` becomes a thin wrapper; its behaviour is unchanged and `display-text-cleaner.spec.ts` is the guard on that (still green, untouched).

Sites swept: `voice-output-stripper.ts`, `voice.service.ts` (was a hand-rolled copy of the display cleaner), `voice-response-condenser.service.ts`, `copilot/services/patch-executor.service.ts`.

## Path 2 — the structured-template branch skipped voice post-processing

`chat.service.ts` returned early on the structured-template branch, before the `channel === 'voice'` block further down, so a voice request routed to a template returned raw markdown and **intact** `[citation:…]` markers. The branch now runs the same `deduplicateResponse()` + `stripForVoice()` as the main flow. The text persisted to the database stays raw, so evaluation still sees the markers.

Belt and braces: `voice.service.ts` and `voice-response-condenser.service.ts` sit downstream of *every* `chat.service` return, and both now clean unconditionally. The condenser is the last code before `tts.synthesize(ssml)`, so nothing that skips an upstream step can leak into speech.

## Changed files

| File | Change |
|---|---|
| `apps/api/src/common/text-cleaning.ts` | **new** — shared hardened patterns: `stripCitationMarkers`, `stripCitationDebris`, `stripResidualMarkdownForSpeech`, `cleanForSpeech` |
| `apps/api/src/common/text-cleaning.spec.ts` | **new** — 21 tests on the primitives |
| `apps/api/src/modules/chat/display-text-cleaner.ts` | patterns moved out; delegates. Behaviour unchanged |
| `apps/api/src/modules/chat/voice-output-stripper.ts` | uses shared strip + final markdown sweep |
| `apps/api/src/modules/chat/voice-output-stripper.spec.ts` | **+5 tests** — Path 1 regressions |
| `apps/api/src/modules/chat/chat.service.ts` | Path 2 — template branch honours `channel: "voice"` |
| `apps/api/src/modules/chat/chat.service.spec.ts` | **+6 tests** — Path 2 regressions, driving the template branch |
| `apps/api/src/modules/voice/voice.service.ts` | uses `cleanResponseForDisplay` instead of a copy of it |
| `apps/api/src/modules/voice/services/voice-response-condenser.service.ts` | shared strip + markdown sweep before the sentence split |
| `apps/api/src/modules/voice/services/voice-response-condenser.service.spec.ts` | **new** — 11 tests at the TTS boundary |

## Commands and results

```
cd apps/api && npx jest
Test Suites: 49 passed, 49 total
Tests:       939 passed, 939 total
```

Baseline on `main` is 48 suites / 905 tests. +1 suite, +34 tests. No test was deleted, skipped, or reworded.

```
cd apps/api && npm run build     # nest build, clean
```

## Verified vs assumed

**Verified**

- The regression tests fail against the pre-fix patterns. Run directly against the regexes on `origin/main`:
  - unterminated marker → `Biomarker tests guide treatment [citation:kb_en_nci_..._v1:kb_` (KB id spoken)
  - two markers → `""` (sentence silently deleted)
  - `**Call 112 if you have chest pain` → `**` survives to TTS
  - `###Next steps` → `###` survives to TTS
- The Path 2 test asserts the structured-template branch is the one that runs (`planExecutor.execute` called) before asserting on its output, so it cannot pass by taking a different branch.
- A Path 2 test asserts the persisted message text still contains `[citation:` — the fix does not damage evaluation data.
- A Path 2 test asserts the **non-voice** channel is unchanged (still cleaned at the controller by PR #82's cleaner), so this is not a behaviour change for web.
- Devanagari is preserved through both the marker strip and the markdown sweep (tested in two specs). The debris rules match punctuation characters directly rather than using `\b`, which is ASCII-only.
- The TTS wiring is a real choke point: `voice.service.ts:206` → `condenser.condense()` → `tts.synthesize(ssml)`, and `voice.controller.ts:117` takes the same condenser.

**Assumed / not verified**

- No live TTS call was made. Verification is at the text boundary — the string handed to `synthesize()` — not on synthesised audio.
- The voice eval (`eval/cases/voice/`, `voice-transcript-eval.ts`) was **not** run. Issue #87's suggestion 3 was to add a `must_not_mention` entry for `[citation:` there; that is not in this PR — the assertions live in Jest instead, where they run on every CI build rather than only on an eval run. Worth a follow-up if the eval surface is wanted too.

## Deliberately not changed

- `copilot/services/eval-bridge.service.ts:49` and `chat/output-verifier.service.ts:343` carry the same loose `\[citation:[^\]]+\]` pattern. Issue #87 names eval-bridge, but both use it only to **count** markers for scoring/verification — they strip nothing and reach no user surface, so neither is a leak path. Changing what they count would change scoring behaviour, which is outside #87's scope.
- No prompt template was edited. No marker leak was traced to prompt wording during this work; if one is found later it needs an SCCF-reviewed PR, per AGENTS.md §1.3.

## Human decisions required

None for merge. One judgement call worth a look: `stripResidualMarkdownForSpeech` removes any surviving `*` and `` ` `` outright rather than trying to infer intent. For spoken text that is the fail-closed direction — a synthesiser has no way to render either — but it does mean a literal asterisk in source text would not be spoken. No KB content relies on that.

### DIFF
--- a/apps/api/src/common/text-cleaning.spec.ts
+++ b/apps/api/src/common/text-cleaning.spec.ts
@@ -1,0 +1,151 @@
+/**
+ * Tests for the shared text-cleaning primitives (issue #87).
+ *
+ * These are the patterns PR #82 hardened for the display surface and that the
+ * voice/TTS surface used to carry weaker copies of. Every case here failed
+ * against at least one of those copies.
+ *
+ * No clinical or safety wording is asserted on — this is machine identifiers
+ * and markup only.
+ */
+
+import {
+  cleanForSpeech,
+  stripCitationMarkers,
+  stripResidualMarkdownForSpeech,
+} from "./text-cleaning";
+
+const KB_ID = "kb_en_nci_types_breast_diagnosis_breast_cancer_biomarker_tests_v1";
+
+describe("stripCitationMarkers", () => {
+  it("removes a complete marker", () => {
+    expect(stripCitationMarkers(`Chemotherapy is a treatment [citation:${KB_ID}:kb_3].`)).toBe(
+      "Chemotherapy is a treatment .",
+    );
+  });
+
+  it("removes an UNTERMINATED marker — the #68/#87 fail-open gap", () => {
+    // The generation stopped inside the marker, so no `]` ever arrived. Every
+    // pre-fix voice pattern required one and spoke the raw KB id aloud.
+    const truncated = `Biomarker tests can guide treatment [citation:${KB_ID}:kb_`;
+
+    const out = stripCitationMarkers(truncated);
+
+    expect(out).not.toContain("[citation:");
+    expect(out).not.toContain(KB_ID);
+    expect(out).not.toContain("kb_en_");
+    expect(out).toBe("Biomarker tests can guide treatment ");
+  });
+
+  it("stops an unterminated marker at the first space, keeping the prose after it", () => {
+    const out = stripCitationMarkers(`Rest well [citation:${KB_ID} and drink water.`);
+
+    expect(out).not.toContain("[citation:");
+    expect(out).toContain("and drink water.");
+  });
+
+  it("does NOT swallow the sentence between two markers (the #82 data-loss bug)", () => {
+    // `[^\]]*` did not exclude `[`, so an unterminated marker matched forward
+    // into the next complete one and deleted the prose in between.
+    const input = "[citation:doc1 Radiotherapy is given daily. [citation:doc2:chunk2]";
+
+    const out = stripCitationMarkers(input);
+
+    expect(out).toContain("Radiotherapy is given daily.");
+    expect(out).not.toContain("[citation:");
+  });
+
+  it("removes a truncated `[citation:` prefix at end of text", () => {
+    expect(stripCitationMarkers("Talk to your oncologist. [cita")).toBe(
+      "Talk to your oncologist. ",
+    );
+  });
+
+  it("removes [source:...] markers in all three shapes", () => {
+    expect(stripCitationMarkers("A [source:nci_breast_v1] B")).not.toContain("[source:");
+    expect(stripCitationMarkers(`A [source:${KB_ID}`)).not.toContain(KB_ID);
+    expect(stripCitationMarkers("Trailing [sou")).toBe("Trailing ");
+  });
+
+  it("removes numbered refs and a truncated trailing one", () => {
+    expect(stripCitationMarkers("Point one [1] and two [23].")).toBe("Point one and two.");
+    expect(stripCitationMarkers("Point one [12")).toBe("Point one");
+  });
+
+  it("removes a raw Sources section and a dangling Sources header", () => {
+    expect(
+      stripCitationMarkers(`Body text.\n\n**Sources:** [citation:${KB_ID}:kb_1] [citation:d:c]`),
+    ).toBe("Body text.");
+    expect(stripCitationMarkers("Body text.\n\n**Sources:**")).toBe("Body text.");
+  });
+
+  it("leaves ordinary bracketed prose alone", () => {
+    const prose = "Chemotherapy (also called chemo) uses drugs [not radiation] to treat cancer.";
+    expect(stripCitationMarkers(prose)).toBe(prose);
+  });
+
+  it("handles empty and undefined input without throwing", () => {
+    expect(stripCitationMarkers("")).toBe("");
+    expect(stripCitationMarkers(undefined as unknown as string)).toBeUndefined();
+  });
+});
+
+describe("stripResidualMarkdownForSpeech", () => {
+  it("removes unpaired ** that survived a bold unwrap", () => {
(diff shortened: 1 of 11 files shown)


gautamgauri (commented):
Reviewed the current diff. The shape is good: the hardened citation stripping is now shared rather than copied, the structured-template voice branch no longer bypasses voice cleanup, and raw persisted text is preserved for evaluation/audit.

I don't see a merge blocker here.

One non-blocking follow-up: add the `[citation:` leak case to the voice eval lane as well as Jest. The unit coverage is strong, but an end-to-end voice-surface assertion would protect against a future wiring bypass outside these helpers.

PR #107, opened 2026-09-09
fix(eval): make the case manifest count what runs, not what parses (#89)

gautamgauri:
Refs #89.

The manifest reported **601 cases**. It counted every case it could *parse*, which is not the same as every case a runner can *execute*. Nine journeys — including caregiver crisis escalation and emergency red flags — read as coverage while running nowhere.

## What this PR does NOT do, deliberately

**It does not port the 9 journeys.** Porting requires choosing an intent mapping (`PERSONAL_SYMPTOMS` / `EMERGENCY` → which existing rubric?) and pass criteria for symptom-worry and emergency journeys. That is clinical policy requiring SCCF review — AGENTS.md §1.3, and issue #89's own "Decisions needed" 1 and 2. Inventing it would swap a visible gap for an invisible wrong answer on the most safety-critical surface in the product.

A half-port would be worse than none: adding `user_messages` alone flips the file into the "gold" lane, and the run then dies at `No rubric found for intent: PERSONAL_SYMPTOMS` — an honest quarantine turned into a runtime crash.

**No expectation in `phase2_journeys.yaml` was weakened, reworded, or removed.** The only edit to that file is its header comment, updating the case-count arithmetic. The 9 cases and their assertions are byte-identical.

So this implements issue #89's **decision 3**: make the count only claim what is executable.

## What it does

**`eval/scripts/case-lanes.ts` (new)** — one home for the lane rules. They previously existed in two states: `case-schema.test.ts` had its own copy, and `case-manifest.ts` had none and counted everything. That divergence is how the 9 got inside the 601.

**Executability is per case, not per file.** A file with one unrunnable case would otherwise mark all its cases as non-coverage — understating the suite as badly as the old count overstated it.

**`cases:check` enforces it.** The check now fails when the executable count drifts — in either direction, so a port that lands raises the number *on purpose* — and prints:

```
✅ Case manifest OK: 572 executable of 601 cases present across 25 files
⚠️  Case file "tier1/phase2_journeys.yaml": 9 of 9 cases are counted but cannot be executed (9 orphan-schema) — issues #71, #89 — …
```

**CI actually runs the guards now.** The eval jest suite ran *nowhere*: `test:ci` is `node --test ci/`, which only covers the CI status helper. PR #88's lane guard and #70's journey-pattern pins have never executed in CI. Added `npm test` to `eval-tier1.yml`, beside the existing manifest step. It needs no API or judge credentials (`jest.config.js` is unit-tests-only). **Reviewer's call** — if you would rather the nightly eval not gate on unit tests, drop that one step; nothing else in the PR depends on it.

## Second blocker found — reported, not fixed

While verifying "592 executable" I found that number is itself wrong. **20 gold-lane cases name an intent `rubrics/rubrics.v1.json` does not define**, so `Evaluator.getRubric` returns null and the case throws before any scoring — the same blocker as the journeys, hiding in files that look runnable.

Verified against the shipped rubric pack:

```
getRubric("RED_FLAGS_URGENT")  -> NULL -> evaluator throws
getRubric("RED_FLAG_URGENT")   -> found
getRubric("PERSONAL_SYMPTOMS") -> NULL -> evaluator throws
getRubric("EMERGENCY")         -> NULL -> evaluator throws
```

| Intent | Cases | Files |
|---|---|---|
| `RED_FLAGS_URGENT` (plural) | 10 | `generalinfo/general_info_100.yaml` |
| `NAVIGATION` | 3 | `gold/language_voice.yaml`, `gold/ux_completeness.yaml` |
| `EMOTIONAL_SUPPORT` | 3 | same |
| `SIDE_EFFECTS_GENERAL` | 2 | `gold/retrieval_grounding.yaml`, `gold/ux_completeness.yaml` |
| `OUT_OF_SCOPE` | 2 | `gold/retrieval_grounding.yaml` |

Note the rubric lookup is **not** canonicalised although the `--intent` filter is (`eval/utils/canonicalize.ts`), so `RED_FLAGS_URGENT` is a genuinely different, undefined intent.

Not fixed here: all three candidate fixes — add the rubrics, canonicalise the lookup, or re-point the cases — change what those cases assert, and 10 of them are red-flag cases. Eval-owner + SCCF call. Recorded as **`docs/RELIABILITY_BACKLOG.md` item 12** and pinned by a test that lists the unmapped intents by name, so the list cannot grow silently.

## Honest headline

| | before | after |
|---|---|---|
| present | 601 | 601 |
| **executable** | *not tracked* | **572** |
| blocked | *not tracked* | 29 — 9 orphan-schema, 20 missing-rubric |

## Changed files

| File | Change |
|---|---|
| `eval/scripts/case-lanes.ts` | **new** — lane + blocker classification, single source of truth |
| `eval/scripts/case-manifest.ts` | manifest v2: `executableCases`, `unexecutableCases`, `blockers`, per-file lanes; drift is an error |
| `eval/scripts/case-manifest.test.ts` | **+7 tests** |
| `eval/scripts/case-schema.test.ts` | lane logic imported instead of duplicated; pins moved to 601/572/29; **+4 tests** |
| `eval/cases/case-manifest.json` | regenerated (`npm run cases:update`) |
| `eval/cases/tier1/phase2_journeys.yaml` | header comment only — **no case, expectation or assertion touched** |
| `.github/workflows/eval-tier1.yml` | run the eval jest suite; manifest step renamed |
| `docs/RELIABILITY_BACKLOG.md` | item 11 corrected 592→572; item 12 added for the 20 |

## Commands and results

```
cd eval && npm test
Test Suites: 7 passed, 7 total
Tests:       92 passed, 92 total          # was 78, none of which ran in CI

cd eval && npm run typecheck              # tsc --noEmit, clean

cd eval && npx ts-node scripts/case-manifest.ts check
✅ Case manifest OK: 572 executable of 601 cases present across 25 files (0 tombstone(s))
```

The drift guard was confirmed by watching it fire: after the rubric blocker was added to the classifier, `check` failed with
`Executable case count changed: manifest records 592, scan finds 572 (of 601 present)` until the manifest was updated. That is the intended behaviour, observed rather than assumed.

`apps/api` is untouched by this branch.

## Verified vs assumed

**Verified — no judge needed**

- **All 601 cases load.** `Evaluator.loadTestCases` parses every one of the 25 files, 0 failures. This is not a YAML problem.
- **0 of the 9 journeys have `user_messages`**, the field `evaluator.ts` passes to `executeConversation` — confirmed by loading the file through the real loader.
- **The rubric lookups return null** for `PERSONAL_SYMPTOMS`, `EMERGENCY` and `RED_FLAGS_URGENT`, against the real `rubrics.v1.json` (output above).
- **Manifest counts round-trip**: `cases:update` then `cases:check` is clean, and the committed manifest is asserted against a fresh scan by test.
- **The guard catches both directions** — a suite silently losing `user_messages`, and an orphan suite being ported — each with its own test.
- **`case-schema.test.ts` behaviour is preserved**: its original four assertions still exist, now reading lane rules from the shared module.

**Assumed / not verified**

- **No Tier1 eval run was attempted.** The judge provider is down (**issue #74**), so pass/fail rates could not be measured. **Tier1 must be re-run after #74 is resolved** to confirm the 572 executable cases in fact execute end-to-end. Everything here is static analysis of loading and lookup, which is exactly the layer the defect lives in — but it is not the same as a green run.
- The claim that a missing rubric aborts the case comes from reading `evaluator.ts` (`getRubric` → `if (!rubric) throw`) plus verifying the lookup returns null. No case was driven through the evaluator to observe the throw, because that needs a live API.
- Voice-lane cases are treated as executable on schema alone; their rubric pack (`voice-rubrics.v1.json`) was not cross-checked the way the gold lane was. If the same defect exists there, this count is still optimistic.

## Human decisions required

1. **The 9 journeys: port or retire?** Needs the intent mapping and pass criteria (issue #89, decisions 1–2). SCCF medical review. Until then the manifest says 572, and `KNOWN_UNRUNNABLE` keeps the file visible with its reason.
2. **The 20 missing-rubric cases** (backlog item 12) — add rubrics, canonicalise the lookup, or re-point the cases. 10 are red-flag cases, so this is not a typo fix.
3. **Should the nightly eval gate on eval unit tests?** Included here; drop that one step if not.

### DIFF
--- a/.github/workflows/eval-tier1.yml
+++ b/.github/workflows/eval-tier1.yml
@@ -101,10 +101,22 @@ jobs:
         working-directory: eval
         run: npm run test:ci
 
-      - name: Verify case manifest (disappearance guard)
+      - name: Eval framework unit tests
+        working-directory: eval
+        # `test:ci` above is `node --test ci/` — it runs the CI status helper
+        # only, so the jest suite under scripts/, runner/ and cases/ never ran
+        # anywhere (issue #89). That includes the lane guard PR #88 added and
+        # the journey-pattern pins from #70: guards nothing executes are not
+        # guards. No live API or judge credentials are needed — jest.config.js
+        # is unit-tests-only.
+        run: npm test
+
+      - name: Verify case manifest (disappearance + executability guard)
         working-directory: eval
         # Hard-fails if any eval case was removed without an explicit tombstone
-        # entry in cases/case-manifest.json (issue #48).
+        # entry in cases/case-manifest.json (issue #48), or if the number of
+        # cases a runner can actually EXECUTE drifts from what the manifest
+        # records (issue #89).
         run: npx ts-node scripts/case-manifest.ts check
 
       - name: Create reports directory
--- a/docs/RELIABILITY_BACKLOG.md
+++ b/docs/RELIABILITY_BACKLOG.md
@@ -516,4 +516,22 @@ more, and should be decided together with QA0904-1.
    schema (which requires deciding the pass criteria for symptom-worry and
    emergency journeys — SCCF medical review per AGENTS.md §1.3) or retire the
    file with a manifest tombstone. Until then, quote executable coverage as
-   592, not 601.
+   572, not 601 — see item 12; the case manifest now reports the figure
+   directly (`npm --prefix eval run cases:check`, issue #89).
+12. P1-11 (found 2026-09-09 while fixing #89): **20 gold-lane eval cases name
+   an intent `rubrics/rubrics.v1.json` does not define**, so
+   `Evaluator.getRubric` returns null and the run throws before scoring —
+   the same blocker as the journeys, in files that otherwise look runnable.
+   Verified against the shipped rubric pack: `RED_FLAGS_URGENT` → null,
+   `RED_FLAG_URGENT` → found. The unmapped intents are `RED_FLAGS_URGENT`
+   (10 cases, `generalinfo/general_info_100.yaml`), `NAVIGATION` (3),
+   `EMOTIONAL_SUPPORT` (3), `SIDE_EFFECTS_GENERAL` (2), `OUT_OF_SCOPE` (2),
+   across `gold/language_voice.yaml`, `gold/retrieval_grounding.yaml` and
+   `gold/ux_completeness.yaml`. Note the rubric lookup is **not**
+   canonicalised, although the `--intent` filter is
+   (`eval/utils/canonicalize.ts`), so a plural is a different intent.
+   Decision needed: add the missing rubrics, canonicalise the lookup, or
+   re-point the cases at existing intents. All three change what those cases
+   assert — including red-flag ones — so this is an eval-owner + SCCF call,
+   not an agent fix. Pinned by `eval/scripts/case-schema.test.ts` so the list
+   cannot grow silently.
--- a/eval/cases/case-manifest.json
+++ b/eval/cases/case-manifest.json
@@ -1,6 +1,6 @@
 {
-  "version": 1,
-  "generatedAt": "2026-07-05T08:16:59.853Z",
+  "version": 2,
+  "generatedAt": "2026-09-09T17:00:50.463Z",
   "totalCases": 601,
   "files": {
     "generalinfo/general_info_100.yaml": {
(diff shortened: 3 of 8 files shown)


gautamgauri:
**CI note — the red check is pre-existing, not this PR.**

`Run Tier1 Retrieval Quality Eval` fails at its last step, `Enforce evaluation status`:

```
3 of 21 eval cases failed. This is a true evaluation failure (not a notification or reporting issue).
```

That workflow has failed on **every** run it has had, including the nightly on `main` (schedule, 2026-09-09 07:17Z) and every recent PR. Nothing on this branch touches retrieval, the API, or any case's expectations.

The two steps this PR is responsible for both **passed** in that same run:

```
success   Eval framework unit tests                                   <- new step
success   Verify case manifest (disappearance + executability guard)
```

So the new `npm test` step is confirmed working in real CI, and the executability guard passes against the committed manifest.

The other two checks are green:

```
API unit tests           pass
Build + config parity    pass
```

PR #113, opened 2026-09-10
fix(web): render only the escalation block in the emergency banner (#111)

### LINKED ISSUE #111: Emergency banner renders raw markdown and duplicates the entire answer (escalated path)
## Summary

On the emergency path (`safety.classification: red_flag` / `self_harm`) the safety banner is handed the **entire** assistant response and renders it as **plain text**, while the message bubble renders the same text through `ReactMarkdown`. A patient on the highest-stakes path therefore sees (a) raw markdown markers and (b) the whole answer twice.

## Evidence — live prod, 2026-09-10, Playwright/Chromium (headless, mobile 412×915)

One `/chat` POST returning `{"classification":"red_flag","actions":["show_emergency_banner","end_conversation"]}`; DOM read from the live page in the same session:

```
banner ("Seek Emergency Medical Care" card): literal '**' = true,  <strong> = 0, <em> = 0
message bubble ([aria-label="assistant message"]): literal '**' = false, <strong> = 8, <em> = 1
```

What the patient actually reads in the banner:

> …Please seek emergency medical care now. **Call for help immediately:** • **112** (national emergency number) or **108** (ambulance service) • **102** (medical emergency helpline, available in many states) • Or have someone drive you…

Bullets and newlines collapse too, so the whole reply (escalation + "Information from trusted sources" + RAG answer + citation markers + disclaimer, ~3,000 chars in the observed case) is one dense red paragraph — and then the identical text is shown again, formatted, in the bubble below it.

## Cause (code refs, unmodified `main`)

- `apps/web/src/components/ChatInterface.tsx:163-166` — `setSafetyBanner({ classification, message: response.responseText })` passes the **full** response text, not the escalation.
- `apps/web/src/components/SafetyBanner.tsx:23` — `<div style={styles.message}>{message}</div>`; no markdown renderer on this component.
- `apps/web/src/components/MessageList.tsx:53` — the bubble path is `ReactMarkdown`, which is why only the banner leaks markup.

The banner styles (small font, bold title, "Seek Emergency Medical Care") read as intended for a short notice, so passing the whole answer looks unintended.

## Impact

Emergency escalations are the one path where the patient most needs to parse the instruction instantly (call 112/108). Raw `**` + a collapsed 3k-char paragraph + a duplicate answer works against that, and it is the first thing rendered when the modal is dismissed.

## Suggested fix shape

UI-only: pass the escalation block alone to the banner (e.g. text before the `**Information from trusted sources:**` join at `chat.service.ts:431-432`), render it with the same markdown pipeline as the bubble, and stop duplicating the full answer in the card. No change to escalation wording or classification logic — prompt/wording edits stay SCCF-reviewed per AGENTS.md §1.3.

Repro artifacts: `~/.hermes/suchi-daily/runs/2026-09-10T19-02-52_seed444043013/` (responses.json, shots/, report.md) + `~/.hermes/suchi-daily/probe_urgent_render.mjs`.

gautamgauri:
Fixes #111.

## What live prod did

Browser QA run `2026-09-10T19-02-52_seed444043013`, question q05 (`messageId f35aa46a`), against `chat.suchitracancercare.org`. One escalated message, two surfaces, DOM read (not a screenshot):

```
banner : literal_asterisks=true,  <strong>=0, <em>=0
bubble : literal_asterisks=false, <strong>=8, <em>=1
```

`ChatInterface.tsx:163-166` set `safetyBanner.message = response.responseText` — the **entire** reply — and `SafetyBanner.tsx:23` rendered `{message}` as plain text. So a patient on the emergency path read `**Call for help immediately:** • **112** (national emergency number) or **108**…` as literal markup collapsed into one dense paragraph, **and** saw the same ~3,000-character answer a second time, correctly formatted, in the bubble below. Highest-stakes path, worst formatting.

## What changed

**API — `apps/api/src/modules/chat/chat.service.ts`**
- Extracted the join string into an exported `ESCALATION_RAG_SEPARATOR` constant (previously an inline literal in two places, lines 424/432).
- On the urgent path, the escalation block is captured into `escalationText` *before* the RAG half is appended, and returned as `safety.bannerText`. It is therefore a byte-exact prefix of `responseText` — no slicing, no re-wording.
- The other two `show_emergency_banner` returns (emergency fast path, safety-classification branch) set `bannerText` to the full reply, which is what the escalation is on those paths.

**Web**
- `SafetyBanner.tsx` renders through `ReactMarkdown` (with `removeCitationMarkers`), the same pipeline `MessageList.tsx:53` uses for the bubble.
- `ChatInterface.tsx` passes `resolveEscalationText(responseText, safety.bannerText)` instead of the whole reply.
- New `apps/web/src/utils/escalationText.ts`: prefers `bannerText`; falls back to slicing at the separator.
- `services/api.ts`: `safety.bannerText?: string`.

### Client-side fallback — the tradeoff

The banner prefers the server field, but keeps a client-side slice at the separator. Without it the fix would do nothing against the currently-deployed API (which has no `bannerText`), since web and API deploy separately. The slice is defensive: no match → returns the response unchanged, so the worst case is today's behaviour, now rendered as markdown. The separator constant is duplicated byte-for-byte on both sides with a comment on each pointing at the other.

### Not changed

Escalation copy is byte-identical (asserted against `ResponseTemplates.S2`), classification and `actions` untouched, safety module not reordered or bypassed. This changes *where* text renders, not what it says.

Residual, deliberate: the escalation block itself still appears in both the banner and the bubble (the bubble is the full transcript record). What is gone is the duplication of the whole ~3,000-char answer. Narrowing that further would mean editing the delivered message text, which is SCCF territory.

## Tests

Regression tests fail on the pre-fix sources — verified by checking the three source files back out to `HEAD~1` and re-running:
- web: **7 failed** (the 2 that passed are the classification-title tests, unrelated to rendering)
- api: the spec fails to compile (`ESCALATION_RAG_SEPARATOR` does not exist)

With the fix:

```
$ cd apps/web && npx vitest run
Test Files  12 passed (12)
     Tests  170 passed (170)          # baseline on origin/main: 155

$ cd apps/web && npx tsc --noEmit     # exit 0

$ cd apps/api && npx jest
Test Suites: 53 passed, 53 total
Tests:       979 passed, 979 total    # baseline on origin/main: 52 suites / 975 tests
```

New coverage:
- `SafetyBanner.test.tsx` — banner text contains no literal `**` and no raw `---`; `<strong>` present; `[citation:` markers do not leak.
- `ChatInterface.escalation.test.tsx` — banner holds only the escalation; the appended answer appears exactly once on the page; markdown renders in both surfaces; the no-`bannerText` fallback path is covered.
- `escalationText.test.ts` — separator slicing, blank/absent `bannerText`, copy left byte-identical.
- `chat.service.escalation-banner.spec.ts` — `bannerText` equals `ResponseTemplates.S2(...)` exactly and is a prefix of `responseText`; the urgent path still escalates and still appends the trusted-sources half; fast path also sets it.

## Unverified assumptions

- Not re-run against live prod — this branch is not deployed. The DOM evidence above is from prod before the fix; the after-state is asserted in jsdom.
- `bannerText` is additive and optional, so an un-upgraded web client is unaffected. Not exercised against a real deployed API pair.

### DIFF
--- a/apps/api/src/modules/chat/chat.service.escalation-banner.spec.ts
+++ b/apps/api/src/modules/chat/chat.service.escalation-banner.spec.ts
@@ -1,0 +1,253 @@
+import { Test, TestingModule } from "@nestjs/testing";
+import { ChatService, ESCALATION_RAG_SEPARATOR } from "./chat.service";
+import { PrismaService } from "../prisma/prisma.service";
+import { AnalyticsService } from "../analytics/analytics.service";
+import { SafetyService } from "../safety/safety.service";
+import { RagService } from "../rag/rag.service";
+import { LlmService } from "../llm/llm.service";
+import { EvidenceGateService } from "../evidence/evidence-gate.service";
+import { CitationService } from "../citations/citation.service";
+import { AbstentionService } from "../abstention/abstention.service";
+import { IntentClassifier } from "./intent-classifier";
+import { TemplateSelector } from "./template-selector";
+import { StructuredExtractorService } from "./structured-extractor.service";
+import { ResponseValidatorService } from "./response-validator.service";
+import { GreetingFlowService } from "./greeting-flow.service";
+import { EmpathyDetector } from "./empathy-detector";
+import { PatientStateService } from "./patient-state.service";
+import { ClinicalKeywordEnforcerService } from "../llm/clinical-keyword-enforcer";
+import { RetrievalToolService } from "../rag/retrieval-tool.service";
+import { QueryDecomposerService } from "../rag/query-decomposer.service";
+import { CrossLingualService } from "../rag/cross-lingual.service";
+import { ExecutionPlannerService } from "./execution-planner.service";
+import { PlanExecutorService } from "./plan-executor.service";
+import { OutputVerifierService } from "./output-verifier.service";
+import { ReviewService } from "../review/review.service";
+import { ObservabilityService } from "../observability/observability.service";
+import { ResponseTemplates } from "./response-templates";
+
+/**
+ * Regression: the chat response must expose the escalation block on its own, so
+ * the web emergency banner can render *only* that block (issue #111).
+ *
+ * Live browser QA on 2026-09-10 (run `2026-09-10T19-02-52`, q05) found the
+ * banner being handed the entire ~3,000-character reply and rendering it as
+ * plain text. DOM proof from the same message:
+ *
+ *   banner : literal_asterisks=true,  <strong>=0, <em>=0
+ *   bubble : literal_asterisks=false, <strong>=8, <em>=1
+ *
+ * The escalation copy is unchanged by this fix — `bannerText` is a byte-exact
+ * prefix of `responseText`.
+ */
+describe("ChatService — escalation banner text (issue #111)", () => {
+  let chatService: ChatService;
+
+  /** The half the LLM contributes on the urgent path. */
+  const RAG_ANSWER =
+    "**Educational answer**:\nA new lump should be *evaluated* by a doctor within a few days.";
+
+  const SESSION = {
+    id: "test-session",
+    createdAt: new Date(),
+    channel: "web",
+    locale: "en",
+    userType: null,
+    status: "active",
+    userContext: "general",
+    cancerType: null,
+    greetingCompleted: true,
+    emotionalState: "neutral",
+  };
+
+  const CHUNK = {
+    id: "chunk-1",
+    docId: "doc-1",
+    chunkId: "chunk-1",
+    text: "A new breast lump should be evaluated by a clinician.",
+    similarity: 0.9,
+    document: { sourceType: "NCI", isTrustedSource: true, title: "Breast changes" },
+  };
+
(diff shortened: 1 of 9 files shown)


gautamgauri (commented):
**Looks ready from this review.** The API exposes the escalation block explicitly, the web renders it through the same Markdown path as the message bubble, and the backward-compatible separator fallback is a sensible deploy-order safeguard.

One merge-order note: #113 and #114 both touch the urgent composition path in `chat.service.ts`. I would merge this one first, then rebase #114 on top of it and resolve that overlap once the reconciler blockers there are fixed. No additional blocker found here.

gautamgauri:
Review status: **READY.**

Re-checked against current `main`: API + web CI are green, the API field is additive/optional, the web fallback is defensive, and the change is presentation/composition only — no escalation copy, classification, or safety policy changes.

I do not see a merge blocker here. This materially improves the highest-risk UI path by preventing the whole ~3k-character answer from being duplicated into a plain-text emergency banner while preserving the transcript in the normal message bubble.

PR #114, opened 2026-09-10
fix(chat): stop the appended answer contradicting the escalation (#112)

### LINKED ISSUE #112: Escalated replies append an unreconciled RAG answer that can contradict the escalation
## Summary

The urgent/escalation path deliberately composes `escalation template + "\n\n**Information from trusted sources:**\n\n" + RAG answer` into a **single** message (`apps/api/src/modules/chat/chat.service.ts:431-432`). Nothing reconciles the two halves, so the RAG half's own triage advice can directly contradict the escalation in the same reply.

## Observed (live prod, 2026-09-10, browser QA run seed 444043013, q05, `messageId f35aa46a`)

Question: *"I found a new hard lump in my breast that wasn't there last week and it's painful. Is this an emergency?"* → `classification: red_flag`, `actions: [show_emergency_banner, end_conversation]`.

In that one message, after the escalation tells the patient to seek emergency care now and lists 112/108, the appended answer says:

> "It's important to see a doctor promptly, ideally within a few days to a week… While a new lump is concerning and needs evaluation, it does **not** typically require an emergency call to 112 or 108 unless you have other severe, life-threatening symptoms."

So the patient is told to call 112/108 immediately and, three paragraphs later, told the call is not required.

## Not deterministic

Re-running the identical question returned a correct escalation with no contradicting line — the appended answer is LLM-generated per turn. So this is an intermittent composition defect, not a stable template bug; it needs a rule, not a one-off edit.

## Why it matters

The escalation half is the safety-critical instruction; the appended half is model-generated triage advice re-derived from retrieved chunks. Letting the second silently override the first in the same message is the failure mode. Note this is the inverse of the pattern in #110 (transport failures scored as failures): here the safe instruction is present and then diluted.

## Suggested fix shape (for the delegation, subject to SCCF on wording)

Either (a) suppress triage/"when to seek care" guidance from the appended answer when an escalation has already fired, or (b) force the appended half to defer ("your clinician will decide how urgent this is") rather than restate urgency thresholds. Option (b) touches answer wording → SCCF-reviewed separate PR per AGENTS.md §1.3; option (a) is composition logic.

Repro artifacts: `~/.hermes/suchi-daily/runs/2026-09-10T19-02-52_seed444043013/` (`responses.json`, `report.md`).

## Related

- #81 — Hinglish red-flag miss (still reproducing; see comment added 2026-09-10)
- New issue filed alongside this one: **#111** — emergency banner renders raw markdown + duplicates the full answer
- Policy question for SCCF/your call, not a defect: a new painful breast lump trips `red_flag` while the bot's own answer says it is not an emergency — false-positive rate on this path

gautamgauri:
Fixes #112.

## What live prod did

`chat.service.ts:431-432` composes one message on the urgent path:

```
S2 escalation template + "\n\n**Information from trusted sources:**\n\n" + RAG answer
```

Nothing reconciles the halves. Browser QA run `2026-09-10T19-02-52_seed444043013`, q05 (`messageId f35aa46a`) caught the consequence in one message:

> call 112/108 **now** … *(escalation)*
> "…it does **not** typically require an emergency call to 112 or 108 unless you have other severe, life-threatening symptoms." *(appended half)*

Non-deterministic — a re-run of the same question escalated cleanly. The stub in the integration test makes it deterministic.

## The rule chosen

**Once an escalation has fired, the escalation block is the only voice on emergency-level urgency.**

New `apps/api/src/modules/chat/escalation-reconciler.ts` removes sentences that make an emergency/triage statement from the **appended half only** — regardless of polarity:

- a reassuring one ("does not require an emergency call") contradicts the escalation directly above it;
- a reinforcing one ("call 108 if you develop a fever") is already printed verbatim in the escalation block, so nothing is lost to the patient.

Polarity-blind matters: it removes the need to parse negation, which is the brittle part. The matcher keys on emergency *action* markers (112/108/102, ambulance, emergency room/department/care/call, "is/isn't an emergency", "urgent care", "go to the nearest hospital", "immediate medical attention"). Descriptive vocabulary is deliberately excluded — "sepsis is a life-threatening infection" is information, not a competing instruction, and survives.

### Why this and not the alternatives

- **Drop the whole appended half on escalation** — too broad. It throws away the grounding and citations the urgent path deliberately retrieves, on every red_flag turn.
- **Drop the whole "what to do next" section** — removes the useful numbered steps along with the one bad sentence.
- **Detect negation specifically** — needs to tell "it does not require 112" from "do not wait, call 112". That distinction is exactly where a regex breaks. Polarity-blind removal sidesteps it.

### Placement

Applied where `urgentRagResponse` is produced, before citation extraction, so citations describe delivered text. If nothing survives, `urgentRagResponse` becomes `null` and the existing template-only path delivers the escalation alone — no dangling "Information from trusted sources:" header.

## Honesty: this is not robust against arbitrary LLM phrasing

It matches known sentence shapes. **A downgrade phrased without any urgency marker survives** — in the very q05 message, "It's important to see a doctor promptly, ideally within a few days to a week" is a timeline-level downgrade with no emergency keyword, and this fix does not remove it. That limit is asserted as an explicit test (`does not catch a downgrade phrased without any urgency marker`) so it is documented rather than assumed away.

Two things make the residual risk acceptable:

1. **Fail-safe in direction.** A miss leaves today's behaviour. A false positive drops a sentence from the generated half, which can never weaken the escalation.
2. **It removes the shape that was actually observed in production**, plus the whole class of explicit emergency restatements.

### 🔶 SCCF review item (AGENTS.md §1.3) — not done here

Closing the remaining class properly means making the generator aware that the turn escalated — i.e. an escalation-aware system prompt, or a policy that the appended half omits "when to seek care" guidance entirely on red_flag turns. Both are prompt/clinical-policy changes and are **not** made in this PR. Flagging for SCCF, alongside the related q05 question already noted in the QA report: a new painful breast lump tripped `red_flag` while the answer said it was not an emergency (fail-safe direction, but the false-positive rate on this path deserves a decision).

## Not changed

Escalation copy, safety keyword lists, safety module ordering, every prompt, `docs/SUCHI_ANSWER_POLICY.md`. This PR can only remove text from the generated half.

## Tests

The three behavioural assertions fail without the wiring — verified by stashing only the `chat.service.ts` change and re-running:

```
✓ delivers the escalation unchanged
✕ does not deliver the contradicting sentence the LLM produced
✓ keeps the non-conflicting part of the appended answer
✕ leaves the emergency numbers to the escalation and the standard disclaimer
✕ falls back to the escalation alone when the whole appended half was triage talk
Tests: 3 failed, 2 passed, 5 total
```

With the fix:

```
$ cd apps/api && npx jest
Test Suites: 54 passed, 54 total
Tests:       991 passed, 991 total     # baseline on origin/main: 52 suites / 975 tests
```

- `escalation-reconciler.spec.ts` (11 tests) — the observed contradicting sentence is removed; surrounding grounded content, list structure and section labels are preserved; a list item that is only triage talk is dropped with its marker; an emptied section label is dropped; descriptive vocabulary is not matched; the known blind spot is asserted.
- `chat.service.escalation-composition.spec.ts` (5 tests) — full pipeline with real `SafetyService`/`AbstentionService`/`IntentClassifier` and a stubbed LLM that emits the contradiction. Asserts the escalation is delivered byte-intact, the contradiction never reaches `responseText`, the useful half survives, and the all-triage case falls back to escalation-only.

No patient data in fixtures: the user text is synthetic and the appended-half fixture is reconstructed prose, not a transcript.

## Unverified assumptions

- **Tier1 eval not run** (needs API access + judge LLM credentials, and this branch is not deployed). Citation confidence should be unaffected: extraction now runs on the sanitised text, and the existing urgent-path citation-repair attaches deterministic citations whenever fewer than 2 survive. Worth a Tier1 run before merge.
- Not re-verified against live prod — this branch is not deployed.
- The reconciler is English-pattern-based. Hindi/Hinglish generated answers restating urgency are **not** covered; only the numerals 112/108/102 would match. Not in scope for #112, worth a follow-up issue.


gautamgauri (commented):
**HOLD before merge.** Two concrete gaps remain in the reconciler:

1. `statesEmergencyTriage()` misses the exact phrase **“emergency medical care”**. A generated sentence such as “This does not require emergency medical care” can therefore survive underneath an S2 escalation even though it directly contradicts the escalation copy. Please cover that phrase shape explicitly and add a regression test.
2. The section-label cleanup currently treats any standalone bold line as a heading. A terminal factual line such as `**Do not stop treatment without speaking to your doctor.**` can therefore be removed as an “empty section,” discarding grounded non-triage content. Restrict heading detection to actual headings and add a terminal-bold-line test.

The overall direction is good—reconcile only the appended half and fail safe by deletion—but these two cases need closing before this sits on the emergency path.

gautamgauri:
Review status: **HOLD — exact observed contradiction is covered, but the emergency-action matcher still has a material gap.**

The reconciler catches the production phrase involving `emergency call to 112 or 108`, and the fail-safe direction is good. But the matcher is not yet broad enough to support the stated invariant that the escalation is the only voice on emergency-level urgency.

Concrete miss: `"You do not need emergency medical care"` is not matched by the current patterns. You match `emergency care`, `medical emergency`, and `immediate medical care`, but not the common phrase **`emergency medical care`**. That sentence could therefore survive underneath `Please seek emergency medical care now.` and recreate the same contradiction without any number or ambulance keyword.

Please add `emergency medical (care|attention|treatment|evaluation)` (and regression tests for both reassuring and reinforcing polarity) before merge. I would also keep the PR's own documented blind spot explicit: timeline downgrades like `see a doctor within a few days` remain outside this regex approach and still need the SCCF-level composition decision.

gautamgauri:
## Review findings addressed — pushed as `8f0c5a8`

Rebased onto `origin/main` first (import-only conflict in `chat.service.ts` around `correctMedicalSpelling` from #125 — both imports kept). All changes are composition logic in `escalation-reconciler.ts`; **no template, prompt, escalation copy or clinical wording was touched.**

### P1 — negated emergency phrasings survived the matcher (Codex + HOLD review item 1)

`emergency` now accepts an optional `medical` before the action noun, and the reverse order `medical emergency` is kept:

```
/\bemergency\s+(?:medical\s+)?(?:room|department|ward|call|care|services?|service|number|helpline|treatment|attention|evaluation|visit|help)\b/i
```

plus an explicit negation-proximity pattern (`not|no|doesn't|does not|isn't|no need|unlikely|…` within 60 chars of `emergency`, in either order) so negated shapes whose noun phrase is not one of the fixed ones are still caught.

Regression tests (`it.each`, both polarity directions) now cover:

- "This does not require emergency medical care."
- "You do not need emergency medical care."
- "There is no need to seek emergency medical care now."
- "This is not an emergency."
- "It's not typically an emergency department visit unless you have other severe symptoms." — the exact live-prod sentence from #112, also asserted as its own case
- "You are unlikely to need emergency medical attention for this."
- "This isn't a medical emergency." / "An ambulance is not needed." / "No need to call 112 or 108 for this." / "This doesn't require emergency care." / "It is not an emergency room visit." / "You don't need emergency services right now."

Each phrase is asserted twice: `statesEmergencyTriage()` returns true, and the sentence is removed from an appended half while the surrounding grounded sentence survives.

### P2 — every bold-only line was treated as a section heading (Codex + HOLD review item 2)

`SECTION_LABEL` is replaced by `isSectionLabel()`. A bold-only line is a heading only if it is ≤ 60 chars and either ends in a colon or matches known section-heading vocabulary (the labels `llm.service.ts` and `response-templates.ts` actually emit: Educational answer, What to do next, Next steps, Key points, Red flags, Questions to ask, …). A full sentence ending in `.` or `।` is never a heading.

Tests: `**Do not stop treatment without speaking to your doctor.**` survives both as a terminal line and as the entire answer (which previously collapsed to empty text and made `ChatService` discard the whole grounded half); a bold Devanagari sentence ending in a danda survives; a long bold prose line survives; and genuine empty headings (`**When to seek emergency care**:`, colon-less `**Red flags**`) are still dropped.

### P2 — bare `\b102\b` stripped legitimate fever guidance (earlier review)

Emergency-number matching now requires a dialling context and rejects measurement readings:

- `call|dial|phone|ring|contact|reach` + optional filler + `112|108|102` (chains such as "112 or 108" included), or
- an `ambulance|helpline|hotline|emergency number/services/line` noun within 24 chars of the number,
- with a `(?!\s*(?:°|degrees?|deg|F|C|mg|ml|%))` guard on every number.

Tests: "Contact your oncology team the same day if your temperature reaches 102°F or higher." survives untouched (and `statesEmergencyTriage` returns false), as do "temperature reaches 102 degrees", "above 102 F", "102°F or higher on an oral thermometer". Still removed: "Call 112 immediately.", "Dial 108 for an ambulance.", "You do not need to call 112 for this.", "It does not typically require an emergency call to 112 or 108.", "Phone the national emergency number 112 without delay.", "108 is the ambulance helpline in Bihar."

### Verification

- `escalation-reconciler` + `chat.service` specs: **6 suites, 82 tests, all passing**
- Full API suite: **60 suites, 1106 passed, 1 skipped**
- `npm run build` clean

### Not changed / still open

- The documented blind spot stands: a **timeline-only downgrade** ("see a doctor within a few days") carries no urgency marker and survives this regex approach. Closing it means telling the generator the turn escalated — a prompt change, and therefore an SCCF decision per AGENTS.md §1.3. The existing test pinning that limit is unchanged.
- No clinical wording was altered anywhere; the reconciler still only ever deletes from the appended half and never touches the escalation block.

gautamgauri:
Re-review on head `8f0c5a8`: **READY**. The prior emergency-composition blockers are closed without changing templates/prompts/clinical wording: `emergency medical care/attention/treatment/evaluation` and related action phrasings are now caught, while bold factual terminal sentences are no longer misclassified as empty headings. The additional tightening around 102/108/112 also avoids stripping measurement text without a dialing context. The documented residual blind spot—timeline downgrades that contain no emergency/urgency marker—remains a separate SCCF composition-policy question, not a blocker for this targeted fix.

Issue #103, opened 2026-09-08
Navigation: rank by nearest *relevant* centre, not by state hierarchy — needs hospital coordinates


gautamgauri:
## Product decision (founder, 2026-09-08)

> State borders do not matter. What is important is to find the **nearest relevant centre — based on the user's needs**.

This supersedes the geographic model currently in `hospital-directory.service.ts`. Recording it as the target design for hospital navigation, with the data work it requires.

## Why the current model is wrong

PR #99 fixes a real bug — before it, a query naming any Bihar district except Patna, Muzaffarpur or Bhagalpur returned **zero** hospitals, so `buildHospitalContextBlock()` emitted nothing and the answer was filled from model prior knowledge (issue #95, the Darbhanga → Patna report).

But its fallback chain is `exact city → same state → adjacent state → full pool`. That is an **administrative hierarchy, not travel burden**, and it was chosen because it is the only thing the data can express. In East India the two diverge badly:

- **Kishanganj and Purnia** are far closer to **Siliguri** (West Bengal) than to Patna. Siliguri is already in the directory.
- **Western Bihar** is closer to **Varanasi** (Uttar Pradesh), which holds **two TMC units** in the directory — `Homi Bhabha Cancer Hospital, Varanasi` and `Mahamana Pandit Madan Mohan Malaviya Cancer Centre`.

So the state rung both over-includes (distant same-state options ranked above nearer cross-border ones) and mislabels — `chat.service.ts:3158` heads the block `--- Regional / Nearby Centres ---`, which becomes false the moment the adjacent-state rung fires.

## The blocker is data, not algorithm

Audited `apps/api/data/hospitals.json` (83 records, and note this path is a **symlink** to `apps/landing/src/content/hospitals.json` — one file, not two):

| Field | Coverage |
|---|---|
| `contact.address` non-empty | **83 / 83** |
| address with street or pincode detail | **64 / 83** (e.g. `Phulwarisharif, Patna - 801505`) |
| **latitude / longitude** | **0 / 83** |
| **district** | **0 / 83** |
| `logistics.nearest_railway` | 53 / 83 |

Nothing in the file can express distance from a patient. `region` is unusable — nine values, and it splits *within* Bihar (HBCH Muzaffarpur is `East India`, AIIMS Patna is `East`). Catchment prose exists on only 2 of 12 Bihar records, and **Darbhanga appears in neither** — so a catchment map would not even have fixed the reported query.

## Target design

[...]

PR #124, opened 2026-09-12
fix(deploy,chat): ship a real hospitals.json in the API image; health reports a missing directory as degraded (#123)

### LINKED ISSUE #123: Hospital directory is empty in production: apps/api/data/hospitals.json is a git symlink that dangles inside the Docker image (every revision since ≤ Sep 1)
## Symptom

Every `suchi-api` revision logs this at startup — confirmed on `00411`, `00413`, `00414`, `00415`, `00416` (Cloud Logging, 2026-09-01 → 2026-09-11):

```
ERROR [WhatsAppNavigatorFlowService] Failed to load hospitals.json from /app/data/hospitals.json: ENOENT: no such file or directory
WARN  [HospitalDirectoryService] { event: "hospital_directory_not_found", path: "/app/data/hospitals.json" }
```

`HospitalDirectoryService` (chat hospital search, `apps/api/src/modules/chat/hospital-directory.service.ts:122-150`) and `WhatsAppNavigatorFlowService` both load from `path.resolve(process.cwd(), "data/hospitals.json")`. On the miss they fall through to an **empty in-memory directory**, so in production:

- `searchHospitals()` returns nothing for every location → `buildHospitalContextBlock()` emits no block → hospital-navigation answers come from the KB/LLM alone, with none of the 83 curated, scored, PMJAY-flagged hospitals.
- The WhatsApp navigator flow has no hospitals at all.
- The Darbhanga/Muzaffarpur defect in #95 and the fallback-chain fix in #99 are being reasoned about against a directory that is **never loaded in prod**; #103 (rank by nearest relevant centre) likewise.

## Cause

`apps/api/data/hospitals.json` is committed as a **symlink** (git mode `120000`) → `../../landing/src/content/hospitals.json`.

- The Docker build context is `apps/api` (`cloudbuild.yaml` `build-api`: `docker build -f apps/api/Dockerfile apps/api`), so the link target is outside the context. `COPY . .` copies the symlink itself; inside the image it points at `/landing/src/content/hospitals.json`, which does not exist → ENOENT.
- The `stage-hospitals` step in `cloudbuild.yaml:13-14` was meant to materialise the file (`cp apps/landing/src/content/hospitals.json apps/api/data/hospitals.json`), but `cp` onto an existing symlink **follows the link and writes the target**; the symlink stays a symlink. The step's own `ls -la` then prints the (dangling-in-image) link and reports success.
- Locally and in Jest the link resolves, which is why every test passes and the two services' comments say "single canonical path".

## Fix (small; needs a deploy to take effect)

1. `cloudbuild.yaml` `stage-hospitals`: replace the symlink with a real file before the Docker build —
   `rm -f apps/api/data/hospitals.json && cp apps/landing/src/content/hospitals.json apps/api/data/hospitals.json` (or `cp --remove-destination`). Mirror in `cloudbuild.gated.yaml`.
2. Make the miss loud instead of a WARN + empty directory: fail startup (or mark `/v1/health` degraded) when `hospitals.json` is absent or parses to zero hospitals. Silent-empty is how this survived ten days across five revisions.
3. Add a check to `scripts/check_deploy_config_parity.py` (or a tiny CI step) that the staged file is a regular file with ≥ 1 hospital.
4. Optional: replace the symlink in git with a real committed copy plus a sync check, so the build does not depend on a staging step at all.

## Verification after deploy

- Startup logs show `hospital_directory_loaded` with `total: 83` (or current count), no ENOENT.
- Prod web: "best cancer hospital near Darbhanga" returns HBCH Muzaffarpur from the directory (this is the #95 acceptance case, which cannot currently pass in prod for this reason).

## Related
- #95, #99, #103 — all assume the directory is loaded in prod
- Found while verifying the #119 deploy (`00416-sw7`); the error is pre-existing and unrelated to that change.

gautamgauri:
Closes #123.

## What was wrong

Production has run with an **empty hospital directory on every revision since at least 2026-09-01**. Startup logs on `00411`…`00416` all show:

```
ERROR [WhatsAppNavigatorFlowService] Failed to load hospitals.json from /app/data/hospitals.json: ENOENT
WARN  [HospitalDirectoryService] { event: "hospital_directory_not_found" }
```

Chain: `apps/api/data/hospitals.json` is a git **symlink** → `apps/landing/src/content/hospitals.json`; `.gcloudignore` excluded `apps/landing/` entirely; so the build upload carried only the symlink; the old `stage-hospitals` step took its "already present (committed copy)" branch and `ls -la`'d the symlink as success; Docker `COPY . .` (context `apps/api`) copied a link that dangles inside the image. Both loaders swallowed the miss and served zero hospitals, while `/v1/health` said `ok`. `cloudbuild.gated.yaml` had **no staging step at all**.

Consequence: #95 / #99 / #103 reason about a directory that is never loaded in prod, and the WhatsApp navigator has no hospitals.

## Changes

| Layer | Change |
|---|---|
| `.gcloudignore` | Exclude `apps/landing` level by level so exactly **one** file, `apps/landing/src/content/hospitals.json`, is uploaded. Verified: `gcloud meta list-files-for-upload` now lists that file and nothing else from landing. |
| `cloudbuild.yaml` `stage-hospitals` | `rm` the symlink, `cp` the canonical file, then **fail the build** unless the result is a regular (non-symlink) file with a `hospitals` array and ≥ 50 ids. No silent fallback branch. |
| `cloudbuild.gated.yaml` | Same staging step added (`build-api` waits for it). Health gate now requires `"status":"ok"` in the body, not just HTTP 200, so a degraded candidate is never promoted. |
| `common/hospital-directory-file.ts` (new) | One loader for both services: path resolution, Windows pseudo-symlink handling (moved from the navigator), throws on missing / dangling / invalid / empty. Plus a status registry the health endpoint reads — no DI coupling between `HealthModule` and `ChatModule`. |
| `HospitalDirectoryService`, `WhatsAppNavigatorFlowService` | Use the shared loader; the miss is logged at **ERROR** with the path and cause; status recorded. Empty fallback kept so the app still boots. |
| `HealthService` | Adds `hospitalDirectory: {loaded, count, path, error, checkedAt}` and returns `status: "degraded"` when the DB is fine but the directory is not loaded. Still HTTP 200 — the Cloud Run startup probe is TCP-only, so nothing restart-loops. |

## Verification

- New `hospital-directory-file.spec.ts` (regular file, pseudo-symlink, ENOENT, **dangling real symlink** — the exact image shape, empty array, missing key) and `health.service.spec.ts` (ok / degraded / error).
- Full API suite: **57 suites / 1048 tests green**; `tsc --noEmit` clean; `scripts/check_deploy_config_parity.py` OK.
- Staging step dry-run on a replica layout: symlink → regular file, `83 hospital ids`, exit 0; with the source removed → `FATAL … missing from build upload`, exit 1.

## After merge

Deploy, then confirm on the new revision: startup log `hospital_directory_loaded` with `total: 83`, `/v1/health` → `status: "ok"` with `hospitalDirectory.loaded: true`, and a prod web ask for a cancer hospital near Darbhanga returns HBCH Muzaffarpur (the #95 acceptance case, which cannot pass today for this reason).

## Notes

- Overlaps `health.service.ts` with #98's `/v1/health/retrieval` work — whichever lands second gets a small rebase; both add fields, neither changes the DB check.
- `navigator-approve.service.ts` resolves `data/hospitals.json` first when appending an approved hospital; in the image that path was a dangling link, so it fell through its candidate list. Worth a look once this ships — not changed here.

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -85,8 +85,16 @@ apps/web/test-results/
 eval/results/
 eval/reports/
 
-# Landing page (separate deployment)
-apps/landing/
+# Landing page (separate deployment) — excluded EXCEPT the hospital directory,
+# which is the canonical source that cloudbuild's `stage-hospitals` step copies
+# into apps/api/data/ for the API image (issue #123). gitignore semantics cannot
+# re-include a file under an excluded directory, so exclude contents level by level.
+apps/landing/*
+!apps/landing/src/
+apps/landing/src/*
+!apps/landing/src/content/
+apps/landing/src/content/*
+!apps/landing/src/content/hospitals.json
 
 
 
--- a/apps/api/src/common/hospital-directory-file.spec.ts
+++ b/apps/api/src/common/hospital-directory-file.spec.ts
@@ -1,0 +1,73 @@
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import {
+  getHospitalDirectoryStatus,
+  readHospitalDirectoryFile,
+  recordHospitalDirectoryStatus,
+  resetHospitalDirectoryStatusForTests,
+} from "./hospital-directory-file";
+
+function tmpCwd(): string {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), "suchi-hospdir-"));
+  fs.mkdirSync(path.join(dir, "data"));
+  return dir;
+}
+
+describe("readHospitalDirectoryFile (issue #123)", () => {
+  it("reads a regular file", () => {
+    const cwd = tmpCwd();
+    fs.writeFileSync(path.join(cwd, "data/hospitals.json"), JSON.stringify({ hospitals: [{ id: "a" }, { id: "b" }] }));
+    const out = readHospitalDirectoryFile(cwd);
+    expect(out.hospitals).toHaveLength(2);
+    expect(out.resolvedVia).toBe("file");
+  });
+
+  it("follows a Windows pseudo-symlink (one-line relative path)", () => {
+    const cwd = tmpCwd();
+    fs.mkdirSync(path.join(cwd, "landing"));
+    fs.writeFileSync(path.join(cwd, "landing/hospitals.json"), JSON.stringify({ hospitals: [{ id: "a" }] }));
+    fs.writeFileSync(path.join(cwd, "data/hospitals.json"), "../landing/hospitals.json");
+    const out = readHospitalDirectoryFile(cwd);
+    expect(out.hospitals).toHaveLength(1);
+    expect(out.resolvedVia).toBe("pseudo-symlink");
+  });
+
+  it("throws on a missing file — the production shape (ENOENT)", () => {
+    const cwd = tmpCwd();
+    expect(() => readHospitalDirectoryFile(cwd)).toThrow(/ENOENT/);
+  });
+
+  it("throws on a dangling real symlink — the exact image shape", () => {
+    const cwd = tmpCwd();
+    fs.symlinkSync("../../landing/src/content/hospitals.json", path.join(cwd, "data/hospitals.json"));
+    expect(() => readHospitalDirectoryFile(cwd)).toThrow(/ENOENT/);
+  });
+
+  it("throws on an empty hospitals array instead of loading nothing silently", () => {
+    const cwd = tmpCwd();
+    fs.writeFileSync(path.join(cwd, "data/hospitals.json"), JSON.stringify({ hospitals: [] }));
+    expect(() => readHospitalDirectoryFile(cwd)).toThrow(/empty/);
+  });
+
+  it("throws when the hospitals key is missing", () => {
+    const cwd = tmpCwd();
+    fs.writeFileSync(path.join(cwd, "data/hospitals.json"), JSON.stringify({ nope: [] }));
+    expect(() => readHospitalDirectoryFile(cwd)).toThrow(/no "hospitals" array/);
+  });
+});
+
+describe("hospital directory status registry", () => {
+  beforeEach(() => resetHospitalDirectoryStatusForTests());
+
+  it("starts as not loaded", () => {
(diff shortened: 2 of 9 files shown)


gautamgauri:
Review status: **READY — recommend merge first in the current queue.**

I re-checked the deploy-path changes against current `main`. The core fix is sound: materialise `apps/landing/src/content/hospitals.json` into the API Docker context as a real file, fail the build if it is absent/empty, expose directory load state in `/v1/health`, and make the gated pipeline refuse a `degraded` candidate. Required API/build-parity CI is green.

This should land before judging #99 in production, because #99 cannot improve hospital routing while the production image has no directory loaded.

One non-blocking hardening suggestion: the build step currently validates/counts with `grep` (`grep -q '"hospitals"'`, `grep -c '"id":'`). That is formatting-dependent and could miscount if the JSON is compacted/reformatted. Prefer a tiny Node/Python JSON parse eventually (`Array.isArray(data.hospitals)` + length >= 50). I would **not block this PR** on that cleanup; the current change is materially safer than production.

gautamgauri:
Hardening suggestion applied in `dd93112`: the `stage-hospitals` step now runs in `node:20-slim` and validates with a real JSON parse (`Array.isArray(d.hospitals) && d.hospitals.length >= 50`) instead of `grep`. Dry-run: 83 hospitals staged from both cloudbuild files; a compacted 3-hospital file fails the build. Parity check still OK.

gautamgauri:
Review update after hardening commit `dd93112`: **READY — previous non-blocking validation concern is resolved.**

The staging step now parses the JSON in Node and asserts a real `hospitals` array with a minimum count instead of relying on `grep`/formatting. That removes the only hardening note from my previous review. The PR is mergeable and required CI is green.

Recommendation unchanged: merge/deploy #124 first, verify `hospitalDirectory.loaded: true` / count 83 on the new revision, then rebase #98 so both hospital-directory and retrieval health fields survive.

PR #125, opened 2026-09-12
fix(chat): chemo-prep template no longer hijacks Hinglish red flags; speech cleanup by input modality, not channel (#115 re-scope of #121)

gautamgauri:
Refs #115. **Re-scope of #121** per its review ("RE-SCOPE before making ready"): only the two deterministic fixes, with the typed-input fix made channel-complete. No safety-module, keyword-list, prompt or patient-facing copy changes. Recommend closing #121 in favour of this PR.

## 1. `selectOutputTemplate`: bare `kya` / `kaise` counted as chemo-*preparation* cues

Any message mentioning chemo plus a Hinglish question word got the "Before Your Chemo Session" checklist — including a caregiver reporting heavy post-chemo bleeding and dizziness (live WhatsApp 2026-09-10 01:35 IST; web QA q01 same day; **still reproduced on revision `00416`** on 2026-09-11). A real preparation cue is now required (`prepare|preparation|ready|day|what to|taiyari|tayari|तैयारी`). Regression test uses the exact live text; positive Hinglish prep cases (`taiyari`, `tayari`) added.

After this the red-flag text falls to the navigate soft-redirect rather than escalating — that is the #81 keyword gap; the SCCF proposal from #121 §4 is now on #81.

## 2. Phase 0 input cleanup: by **modality**, not channel

`cleanVoiceInput` (stutter/repeat/filler removal for Web Speech API output) ran on every channel and rewrote typed Hinglish before classification: `meri didi` → `meri di`, `papa` → `pa`, `bahut bahut` → `bahut`. #121 exempted WhatsApp only; the review pointed out web is typed input too.

| input | cleanup |
|---|---|
| channel `voice` | `cleanVoiceInput` |
| any channel with new optional `ChatDto.inputMode = "voice"` | `cleanVoiceInput` |
| everything else (typed: web, WhatsApp, app) | `correctMedicalSpelling` only |

The web mic path (Web Speech API → sent as channel `web`) now tags its messages: `MessageInput` remembers that recognition wrote into the draft → `onSend(text, { inputMode: "voice" })` → `ChatInterface` → `/v1/chat` `inputMode`. Typed messages send **no flag**, so the API contract and existing callers are unchanged; the flag resets after each send.

## Dropped from #121 (not lost)

The shared `timeout-fallback.ts`, the WhatsApp per-turn bound and the reuse of the web timeout guidance on WhatsApp. Root cause of those timeouts was CPU throttling of the fire-and-forget turn, fixed in #119 and deployed as `00416` (ledger since deploy: 15/15 processed, 0 failed, ~5 s turns). Reusing the guidance copy would also have expanded patient-facing text to a new channel. The work is preserved in `195d7b2` on `fix/115-whatsapp-hinglish-pipeline` if wanted as a separate belt-and-braces PR.

## Verification

- New `chat.service.input-cleanup.spec.ts` (real `SafetyService`/`IntentClassifier`, mocked LLM/RAG/DB): WhatsApp typed text intact; **web typed text intact**; web + `inputMode: "voice"` cleaned; `voice` channel cleaned; typed text still gets medical-spelling fixes; rule-layer gap documented.
- `structured-output-templates.spec.ts`: #115 regression + Hinglish prep positives + bare-question-word negatives.
- `MessageInput.test.tsx`: typed → no flag (backward compatible), mic → `{ inputMode: "voice" }`, flag resets.
- API: **56 suites / 1046 tests** green; web `MessageInput` 19/19; `tsc --noEmit` clean in both apps.
- Line endings preserved (CRLF files kept CRLF) so the diff is the change only: 8 files, +132/−12, plus one new spec.

### DIFF
--- a/apps/api/src/modules/chat/chat.service.input-cleanup.spec.ts
+++ b/apps/api/src/modules/chat/chat.service.input-cleanup.spec.ts
@@ -1,0 +1,235 @@
+import { Test, TestingModule } from "@nestjs/testing";
+import { ChatService } from "./chat.service";
+import { PrismaService } from "../prisma/prisma.service";
+import { AnalyticsService } from "../analytics/analytics.service";
+import { SafetyService } from "../safety/safety.service";
+import { RagService } from "../rag/rag.service";
+import { LlmService } from "../llm/llm.service";
+import { EvidenceGateService } from "../evidence/evidence-gate.service";
+import { CitationService } from "../citations/citation.service";
+import { AbstentionService } from "../abstention/abstention.service";
+import { IntentClassifier } from "./intent-classifier";
+import { TemplateSelector } from "./template-selector";
+import { StructuredExtractorService } from "./structured-extractor.service";
+import { ResponseValidatorService } from "./response-validator.service";
+import { GreetingFlowService } from "./greeting-flow.service";
+import { EmpathyDetector } from "./empathy-detector";
+import { PatientStateService } from "./patient-state.service";
+import { ClinicalKeywordEnforcerService } from "../llm/clinical-keyword-enforcer";
+import { RetrievalToolService } from "../rag/retrieval-tool.service";
+import { QueryDecomposerService } from "../rag/query-decomposer.service";
+import { CrossLingualService } from "../rag/cross-lingual.service";
+import { ExecutionPlannerService } from "./execution-planner.service";
+import { PlanExecutorService } from "./plan-executor.service";
+import { OutputVerifierService } from "./output-verifier.service";
+import { ReviewService } from "../review/review.service";
+import { ObservabilityService } from "../observability/observability.service";
+
+/**
+ * Issue #115 — Phase 0 input cleanup must follow the input MODALITY, not the
+ * channel: speech-recognition output gets stutter/repeat/filler removal, typed
+ * text on any channel (WhatsApp, web) only gets medical-spelling normalisation.
+ * The texts below are the tester's own messages as quoted in the issue (no
+ * patient data).
+ *
+ * The texts below are the tester's own messages as quoted in the issue (no
+ * patient data). All LLM / RAG / DB dependencies are mocked; the point of these
+ * tests is the *control flow* of `handle()` for `channel: "whatsapp"`, not the
+ * answer content.
+ */
+
+const HINGLISH_BENIGN = "breast cancer ke early symptoms kya hote hain? kab tak dhyan dena chahiye";
+const HINGLISH_RED_FLAG =
+  "meri didi chemo ke baad se bahut kamjor hai, aaj bleeding bahut zyada ho gayi aur chakkar aa raha hai. kya karu??";
+
+const CHUNKS = [
+  {
+    docId: "doc1",
+    chunkId: "chunk1",
+    content: "Breast cancer symptoms include lumps, nipple changes, and skin changes.",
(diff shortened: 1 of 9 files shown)


gautamgauri (commented):
Review update: **HOLD — the re-scope is much cleaner than #121, but two edge cases remain before I would merge.**

1. **The chemo-prep matcher is still too broad.** Removing bare `kya|kaise` fixes the reproduced Hinglish case, but `isPrep` still treats bare `day` and bare `what to` as preparation cues:

```ts
/\b(prepare|preparation|ready|day|what to|taiya?a?ri|tayy?ari)\b/i
```

That can still route non-preparation post-chemo questions into `CHEMO_DAY_PREP`, e.g. `I had chemo and now I am vomiting — what to do?` or `one day after chemo I feel very weak`. This is the same failure class in a different phrase shape. Please make the English cues preparation-specific (for example `chemo day`, `day of chemo`, `before chemo`, `what to bring`, `prepare/preparation/getting ready`) rather than matching `day` / `what to` anywhere. Add negative regressions for post-chemo `what to do` and `one day after chemo`, while keeping the existing positive `what to do on chemo day` behaviour through a more specific phrase.

2. **The web-mic flag can still cause typed text to be speech-cleaned.** `usedVoiceRef` becomes true once recognition writes anything into the draft, and then the *entire combined message* is sent as `inputMode: "voice"`. If a user types `meri didi ` and then dictates the rest, the server runs `cleanVoiceInput()` over the typed prefix too, recreating `didi → di`. Similarly, if the user dictates, manually edits/replaces the transcript, and sends, the flag remains `voice`.

The safe rule should be: mark the whole message `voice` only when it is entirely mic-generated and has not subsequently been manually edited. Mixed typed+mic or post-dictation-edited drafts should take the spelling-only path unless we start carrying per-span modality, which would be overkill here. Please add regressions for both mixed typed+mic input and manual edit after dictation.

Everything else about this re-scope looks right: timeout-copy expansion is gone, the #81 safety-keyword decision is separated, typed web/WhatsApp now share the same default path, and CI is green on this head.

gautamgauri:
Two further review findings (from the maintainer's re-review, not yet posted here) — both real, both fixed in `86bfbd6`:

**1. Matcher still had the same bug in English.** After dropping `kya|kaise`, bare `day` and `what to` remained, so *"Chemo ke baad vomiting ho rahi hai, **what to** do?"* and *"**One day** after chemo I feel very weak"* still selected the chemo-day checklist. Cues are now phrase-level only: `prepare|preparation|preparing`, `get(ting)?/be ready`, `ready for|hona|ho`, `chemo day`, `day of chemo`, `before (my|the|first|pehli) chemo`, `chemo se pehle`, `what to bring|carry|wear|eat|expect`, `taiyari|tayari`, `तैयारी`. Both quoted sentences plus a "fever 3 days after chemo" case are regression tests → `null`; four phrase-level positives added.

**2. Whole-draft voice tagging leaked speech cleanup onto typed text.** Typing `meri didi ` then dictating, or dictating then hand-editing, sent the whole message as `inputMode: "voice"`. The draft now tracks provenance `none | pure | mixed`: only an entirely dictated draft with no keystrokes after recognition is `pure` and gets the flag; a non-empty typed prefix at recording start, or any `onChange` after dictation, makes it `mixed` → spelling-only path. Tests: typed-prefix+dictation → no flag; dictation+manual edit → no flag; pure dictation → flag; flag resets after send.

Templates spec 57/57, input-cleanup spec 6/6, MessageInput 21/21, `tsc` clean in both apps.

PR #130, opened 2026-09-12
fix(rag,db): non-blocking FTS restore — plain column + trigger, batched backfill, CONCURRENT index; probe accepts both shapes (#92 follow-up)

gautamgauri:
Refs #92, #98. Follow-up to the **2026-09-12 incident** (recorded in `docs/RELIABILITY_BACKLOG.md`): applying the first version of `20260908000000_restore_kb_chunk_fts` re-added `content_tsv` as a `GENERATED ALWAYS … STORED` column. Postgres rewrote the whole `KbChunk` table under an ACCESS EXCLUSIVE lock and rebuilt every index including the pgvector HNSW one; on `db-f1-micro` it ran 15+ minutes, queued all retrieval, exhausted the pool and took `/v1/health` down until the backend was terminated. Nothing was applied. **No persistent database has recorded this migration** (checked: suchi-db is the only Postgres for this product; its `_prisma_migrations` ends at `20260622`), so the migration is rewritten in place under the same name rather than followed by a forward migration.

## The rollout sequence (per review)

| step | where | what |
|---|---|---|
| 1 | `migration.sql` | `SET lock_timeout='5s'`; add `content_tsv` as a **plain, nullable** tsvector, no default — catalog-only |
| 2 | `migration.sql` | **trigger first**: `BEFORE INSERT OR UPDATE OF content` → `NEW.content_tsv := to_tsvector('simple', NEW.content)` |
| 3 | `scripts/sql/kb_fts_safe_rollout.py` | backfill in **committed batches** (default 1 500, sleep between, pauses when lock waiters appear); never one transaction |
| 4 | script | **after** the backfill: `CREATE INDEX CONCURRENTLY` (drops an INVALID leftover first; outside any transaction) |
| 5 | script | verify `COUNT(*) WHERE content_tsv IS NULL = 0`, `pg_index.indisvalid AND indisready`, live lexical hits > 0 |
| 6 | script | only then `prisma migrate resolve --applied 20260908000000_restore_kb_chunk_fts` |

On a table with ≤ 5 000 rows (dev, CI, the PGlite test) the migration itself does the backfill and a plain index build; on a production-sized table it stops with a `NOTICE` pointing at the script. A legacy STORED generated column (a DB that never ran `20260606`) is left untouched — Postgres maintains it.

## Probe strengthened (`kb-fts.sql.ts`, `KbFtsHealthService`)

Reports **how** the column is maintained — generation expression *or* the enabled trigger and its function body (which carries the `'simple'` config) — and whether the GIN index is **valid**, not merely present (`pg_index.indisvalid AND indisready`; an interrupted concurrent build leaves an invalid index that the planner ignores).

| state | verdict |
|---|---|
| column absent | unavailable |
| plain column, no enabled trigger (what `prisma migrate diff` emits) | unavailable |
| trigger or generated, wrong config | degraded |
| index missing | degraded |
| index present but INVALID | degraded (new) |
| maintained on `'simple'`, index valid | ok |

The runtime check is structural only; the full-table `NULL` count runs in the deployment procedure, not on every `/health`.

## Tests

PGlite (real Postgres, real migrations replayed oldest-first, shipped query): trigger populates on INSERT and on UPDATE OF content · a `db push`-shaped schema is repaired with the trigger and answers queries · unmaintained column → unavailable · **INVALID index → degraded** · the `20260606`-drop and config-mismatch cases unchanged. `kb-fts.spec.ts` 17/17; full API suite **60 suites / 1080 tests**; `tsc` clean; `prisma validate` ok; parity and secret guards ok.

## Runbook

`docs/OPERATIONS_RUNBOOK.md` §4a "Schema changes on `KbChunk` — never a table rewrite": which statements rewrite, the six-step safe pattern, the script invocation, and how to find/terminate a lock holder. Rule 4 under *Deploying* points at it.

## After merge

Run in a night-IST window:
```
export DATABASE_URL="$(gcloud secrets versions access latest --secret=database-url)"
python3 scripts/sql/kb_fts_safe_rollout.py            # dry run
python3 scripts/sql/kb_fts_safe_rollout.py --execute
```
then `GET /v1/health/retrieval` → `fullTextSearch.status: ok`. No code deploy is required for the column to start being used; the running revision already ships the query and the probe.

### DIFF
--- a/apps/api/prisma/migrations/20260908000000_restore_kb_chunk_fts/migration.sql
+++ b/apps/api/prisma/migrations/20260908000000_restore_kb_chunk_fts/migration.sql
@@ -1,4 +1,4 @@
--- Restore full-text search on KbChunk (issue #92).
+-- Restore full-text search on KbChunk (issue #92) — NON-BLOCKING implementation.
 --
 -- HISTORY
 --   20260120163141_add_fts_to_kbchunk  added `content_tsv` as a STORED generated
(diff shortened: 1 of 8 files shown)


gautamgauri:
Safeguard from the review applied in `696f56c`.

**`migration.sql` now either completes or refuses.** With ≤ 5 000 rows it does column + trigger + backfill + plain GIN index (dev, CI, PGlite). With more rows it `RAISE EXCEPTION`s unless the column is already fully populated **and** the GIN index exists and is valid — a DO-block exception rolls the statement back, so nothing is committed and `prisma migrate deploy` (including the gated pipeline's migration job) cannot record it as applied ahead of the backfill. After the rollout script has finished, the same file is a harmless no-op.

**The rollout script bootstraps the schema itself** (`BOOTSTRAP_SQL`, the same column + trigger DDL), then batched backfill → `CREATE INDEX CONCURRENTLY` → verify → `prisma migrate resolve` → re-runs `migration.sql` once to prove it is now a no-op on that database.

**Tests (PGlite, 5 001 rows via `generate_series`):** the migration raises with the *refusing to backfill/index inside a migration* message and leaves no column/trigger behind; after the script's steps it succeeds as a no-op and the probe reports trigger enabled, index valid, zero NULLs. `kb-fts.spec.ts` 19/19; `tsc` clean; `prisma validate` ok. Runbook §4a documents the refusal.

gautamgauri:
Review update on head `696f56c`: **READY**, with one non-blocking hardening/wording note.

The important safeguard from the previous review is now present: on a production-sized table the migration refuses instead of silently returning success, and the rollout script owns bootstrap → batched backfill → concurrent GIN build → verification → `migrate resolve --applied`. The new PGlite 5,001-row case is the right regression to pin that invariant. Required API CI is green.

One nuance: this repo is on Prisma 5.22, and Prisma Migrate does not implicitly wrap an entire PostgreSQL `migration.sql` in one transaction. In this file `CREATE OR REPLACE FUNCTION kbchunk_content_tsv_maintain()` is a statement before the `DO` block. If the large-table guard raises inside the `DO`, the table/trigger work in that `DO` is rolled back, but the preceding function may already have been committed. That is harmless here — the function is unused until a trigger exists and the rollout script replaces it — so I do **not** consider it a merge blocker. But the comments currently say effectively “nothing is committed,” which is stronger than the actual Prisma-5 guarantee.

Two clean options: either wrap the small-table/refusal migration body explicitly in `BEGIN`/`COMMIT` (ordinary GIN build is fine for the <=5k path), or soften the wording to “no KbChunk schema/data change is committed; an unused helper function may remain.”

Operationally, I am comfortable merging this once that nuance is understood. Production execution should still be via `kb_fts_safe_rollout.py` in a low-traffic window, not `migrate deploy`.

gautamgauri:
Took the stronger option in `7356e22`: the whole migration is now a single `DO $do$ … $do$` block with the trigger function created inside it, so on Prisma 5 (no implicit transaction) a raise in the large-table guard rolls back function, column and trigger together — the "nothing is committed" wording is now literally true. The PGlite 5,001-row refusal test additionally asserts `pg_proc` has no `kbchunk_content_tsv_maintain` afterwards. 19/19, `tsc` clean.

gautamgauri:
Production rollout evidence changes my disposition on this PR: **HOLD / redesign around a GIN expression index; do not resume the batched column backfill.**

The safety mechanism worked — no outage or lock pile-up — but the economics of the design are wrong on this instance. Updating ~48k `KbChunk` rows to materialise `content_tsv` creates new heap tuples and index maintenance; measured production cost rose to ~190 ms/row, with ~36.7k rows still NULL. Continuing means hours of write amplification and bloat purely to precompute a value PostgreSQL can index directly.

PostgreSQL supports exactly the simpler shape we need:

```sql
CREATE INDEX CONCURRENTLY kb_chunk_content_tsv_idx
ON "KbChunk" USING GIN (to_tsvector('simple', content));
```

Then the shipped lexical query must use the matching two-argument expression in both the predicate/rank path, e.g. `to_tsvector('simple', c.content) @@ query`. The explicit `'simple'` config matters: PostgreSQL only uses a text-search expression index when the query expression/config matches the indexed expression.

**Important transitional correctness issue:** production currently has a *partially populated* plain `content_tsv` (~12k filled / ~36.7k NULL). `fullTextSearchWithMetadata()` executes `KB_FTS_SEARCH_SQL` directly and calls `recordQuerySuccess()` whenever the SQL succeeds. Now that the column exists, lexical retrieval can therefore succeed over only the populated subset instead of failing closed to vector-only. That is a worse transitional state than the original missing-column state because hybrid scoring can treat incomplete lexical evidence as real coverage.

Recommended cutover:

1. **Do not resume the backfill.**
2. In a low-traffic window, disable the partial-column path: `SET lock_timeout='5s'`; drop the maintenance trigger and drop/rename the plain `content_tsv` column (metadata DDL; fail rather than wait if the lock is unavailable). This restores the known-safe vector-only fallback until the new query ships. Do not rewrite the table to reclaim the dropped-column bytes.
3. Build the expression GIN with `CREATE INDEX CONCURRENTLY` and monitor DB CPU/latency. It still scans/computes over the table, but it does **not** UPDATE every heap row or churn the HNSW index through 48k backfill updates.
4. Rewrite #130 so `KB_FTS_SEARCH_SQL` uses `to_tsvector('simple', c.content)`; remove the column/trigger ownership from Prisma/migration/probe. The health probe should require the expression index to be present, `indisvalid && indisready`, and verify its definition/config (not merely its name).
5. Fresh/small DB migration: create the expression GIN normally. Production rollout script: `CREATE INDEX CONCURRENTLY` outside a transaction, verify with a real lexical query **and `EXPLAIN` showing the expression GIN is used**, then only mark `20260908000000_restore_kb_chunk_fts` applied.
6. Deploy the query/probe and require `/v1/health/retrieval` → FTS `ok`, then replay the retrieval acceptance probes.

This also lets `schema.prisma` stop pretending to own an `Unsupported("tsvector")` column that the application does not need.

One nuance: the stored-column approach can avoid recomputing `to_tsvector` for matched rows and may be marginally faster at query time, but PostgreSQL's own FTS documentation describes the expression-GIN approach as the simpler/lower-storage alternative, and on this 48k-row medical KB the operational cost/risk we just measured outweighs that small advantage.

So I withdraw my earlier READY disposition on the current #130 shape. The incident response was safe; the production measurement is exactly the evidence needed to simplify the design before we commit to it.

gautamgauri:
Redesign done per the HOLD/redesign finding above — **PR #133** (`fix/92-fts-expression-index`).

**Production state now (2026-09-12 ~19:50 UTC):**
- backfill stopped permanently; trigger, function and plain `content_tsv` dropped (catalog-only, 3 s, zero lock waiters);
- `kb_chunk_content_tsv_idx` rebuilt as `GIN (to_tsvector('simple', content))` with `CREATE INDEX CONCURRENTLY` — 90 s, 36 MB, `indisvalid`/`indisready` true;
- verified: definition `to_tsvector('simple'::regconfig, content)`; live search `cancer treatment` → 11,791 rows; `EXPLAIN` → Bitmap Index Scan on the index;
- `20260908000000_restore_kb_chunk_fts` marked applied via `prisma migrate resolve`; the rewritten `migration.sql` re-runs as a no-op.

Health stayed `ok` throughout. The partial-FTS correctness risk is closed: with the column gone, the running revision's `c.content_tsv` query fails cleanly again (vector-only fallback) until #133 deploys the expression query and probe.

#133 rewrites `kb-fts.sql.ts` (expression is the single source), the probe (index present / valid / right expression; legacy column reported), the migration (atomic DO, size guard now on the index), `schema.prisma` (no tsvector field) and the rollout script (concurrent build → verify incl. `EXPLAIN` → resolve). PGlite 20/20 incl. planner-uses-index, no-index → degraded not unavailable, and the 5,001-row refusal.

Issue #126, opened 2026-09-12
WhatsApp: a pregnancy question is answered with breastfeeding/nursing-baby content (drugs in breast milk)


gautamgauri:
## Summary

On the public WhatsApp number, a question about whether cancer treatment affects a baby **in the womb**
is answered with content about **breast milk and nursing infants** — and tells the user that breastfeeding
is stopped — for a woman who is *pregnant*, not nursing. The fetal-risk content **does exist in the KB**
and is retrieved correctly when the same scenario is rephrased, and the *same text* produced a different,
better answer when re-sent 30 minutes later. So this is retrieval ranking / answer composition, not a
content gap.

**Severity: P1.** Wrong-scenario clinical content on a public channel, non-deterministic on identical
input. No red flag was involved, so no emergency was missed.

## Evidence — live public number, 2026-09-12 (all sends confirmed against fresh message ids)

**Probe A — the case, original wording (13:03 IST, bubble `E7787D1018A8357319`)**

> Q: meri mausi ko cancer hai aur wo pregnant hai, kya cancer ki dawai se bachcha affected hoga? exact batao
>
> A: …मैं समझी कि आप जानना चाहते हैं कि अगर आपकी मासी को कैंसर है और वह गर्भवती हैं, तो क्या कैंसर की
> दवाइयों का बच्चे पर असर होगा। गर्भावस्था के दौरान स्तन कैंसर हो सकता है। यदि कैंसर के इलाज के लिए
> कीमोथेरेपी दी जाती है, तो कुछ एंटीकैंसर दवाएं (जैसे साइक्लोफॉस्फेमाइड और मेथोट्रेक्सेट) **स्तन के दूध में
> उच्च स्तर पर मौजूद हो सकती हैं और नर्सिंग शिशु को प्रभावित कर सकती हैं। इसलिए, यदि कीमोथेरेपी दी जानी है,
> तो स्तनपान बंद कर दिया जाता है।**

No fetal risk is mentioned at all: the reply is about a nursing baby, and the actionable line is
"breastfeeding is stopped".

**Probe B — the SAME text, re-sent ~13:25 IST (bubble `3D4AF6174A2D12C9EE`)** — *not self-consistent:*

> A: …शैक्षणिक जानकारी: गर्भावस्था के दौरान कैंसर के इलाज का बच्चे पर असर इलाज के प्रकार और गर्भावस्था के
[...]

Issue #86, opened 2026-09-06
KB index is 34% duplicate rows — 25,308 extra chunks across 528 docs from two overlapping ingest runs


gautamgauri:
## Summary

**34% of the production KB vector index is duplicate content.** Confirmed by direct query against `suchi-db` on 2026-09-06.

This was found while verifying a hypothesis from PR #83 (which fixes the *symptom* — duplicated bullets in chemotherapy answers, issue #67). The agent flagged "stale chunk rows" as a medium-confidence guess it could not check without database access. It is confirmed, and the scale is much larger than the one answer that surfaced it.

## Measured

```
total chunks                    : 73,802
distinct docs                   :  1,439
duplicate groups (same doc,
  byte-identical content)       : 24,895
extra rows beyond the first     : 25,308   <-- 34% of the index
docs affected                   :    528   (of 1,439)
```

Distribution of copies per group:

| copies | groups |
|---|---|
| 2 | 24,785 |
| 3 | 2 |
| 4 | 57 |
| 5 | 2 |
| 6 | 21 |
| 8 | 16 |

## Root cause: two overlapping ingest runs, not a single-run bug

Chunk creation dates:

```
2025-12-31 : 25,065
2026-01-01 :  6,313
2026-01-02 : 42,407
2026-02-17 :     17
```

And decisively:

```
duplicate pairs created within one ingest day : 19
duplicate pairs spanning MORE than one day    : 24,876
```

So essentially every duplicate is one row from the Dec 31 / Jan 1 run and one from the Jan 2 run. A second ingest wrote content that already existed instead of replacing it.

`apps/api/src/scripts/ingest-kb.ts` never prunes stale chunk rows. PR #83 adds pruning for the forward case, but **that does not clean up the 25,308 rows already in production.**

## Why this matters beyond #67

The visible symptom was one chemotherapy answer repeating every sentence. The systemic effects are worse and would not show up as an obvious defect:

1. **Retrieval diversity is degraded across the board.** The reranker uses `topK = 6` (`reranker.service.ts:181`). When duplicates occupy slots, the answer is built on *fewer distinct sources* than intended. Every RAG answer touching one of the 528 affected docs is potentially weaker, not just the ones that visibly repeat.
2. **`chunkId` dedup cannot catch it.** Every merge layer dedupes on `chunkId`, and these are distinct rows with distinct ids carrying identical text — so they pass every existing guard. This is exactly why PR #83 had to add fingerprint-based dedup.
[...]

PR #131, opened 2026-09-12
fix(rag): drop reference-list chunks at retrieval time — PubMed link blocks no longer out-rank the prose that answers (#129)

### LINKED ISSUE #129: Retrieval: PDQ reference-list chunks (PubMed link blocks) are indexed as evidence and can out-rank the prose that answers the question
## What

Found while diagnosing #126. For the Hinglish pregnancy question, the two **top-ranked** chunks (vecSim 0.750 / 0.729) from `kb_en_nci_types_breast_hp_pregnancy_breast_treatment_pdq_v1` are **reference lists**:

- chunk `::25` — 1400 chars, 4 `[[PUBMED Abstract]]` links, 8 URLs
- chunk `::26` — 1400 chars, 3 `[[PUBMED Abstract]]` links, 7 URLs, paper titles like "Fetal outcome after in utero exposure to…"

The chunk that actually answers the question (`### Chemotherapy … safe after the first trimester`, `::18`) ranked **5th**. Paper titles in a reference section embed very close to the queries they are about, so with the lexical arm dead (#92) they routinely win.

Every NCI PDQ document in `kb/en/02_nci_core/pdq/` ends each section with a `###### References` block, so this affects the whole PDQ corpus (528 docs), not one file. Consequences:

- Evidence slots (top-6) are spent on chunks with no answerable prose → thinner answers, more abstentions, and answers composed from whichever *other* prose chunk survives (the #126 breastfeeding drift).
- Citation confidence can be GREEN while the "sources" are link lists.
- Any judge/eval that counts "citations attached" is satisfied by these chunks.

## Fix options

1. **Retrieval-time filter (fast, no re-ingest):** in `rag.service` / the evidence gate, drop or heavily penalise chunks where citation markup dominates — e.g. `PUBMED` link count ≥ 2, or URL characters > 40% of the chunk, or the chunk starts inside a `###### References` block. Ship this first; it takes effect on the next deploy.
2. **Ingest-time fix (clean):** in the NCI ingestion pipeline (`scripts/nci-ingestion`), strip `###### References` sections before chunking, and re-ingest PDQ docs. Pairs naturally with the duplicate-row cleanup in #86 (same re-index).
3. Add a KB health metric: share of indexed chunks that are reference-only, so this cannot silently return.

## Related
- #126 (symptom), #92 / #98 (lexical arm dead), #86 (34% duplicate chunks — same re-index window), #54 (citation policy)

gautamgauri:
Closes #129. Refs #126 (its regression case is in the spec).

## Why

Every NCI PDQ section ends with a `###### References` block: numbered entries with `[[PUBMED Abstract]](url)` links. The ingest chunker indexes those like prose, and paper titles embed very close to the questions they cite. For the pregnancy question in #126 the two top-ranked chunks (vecSim 0.750 / 0.729) were reference lists, the `### Chemotherapy` chunk that answers ranked **5th**, and the model composed from the only other prose chunk it was given (the lactation paragraph).

**Measured on production** (`scripts/sql/kb_reference_chunk_audit.sql`, read-only, same rule as the filter):

| indexed chunks | reference-dominant | share | docs affected | chunks with any PubMed link |
|---|---|---|---|---|
| 48,737 | **11,535** | **23.7 %** | 245 | 12,914 |

All in `02_nci_core`. Nearly a quarter of the evidence pool is link lists. Every one is a candidate the evidence gate and the LLM can be handed, and each one that ranks displaces a prose chunk from the top-6.

## The rule (conservative, explainable)

`reference-chunk-filter.ts` — a chunk is dropped only when it is *dominated* by citation markup. `proseRatio` = characters left after stripping markdown links, URLs **and scheme-less URL fragments** (a chunk that starts inside a link carries `bi.nlm.nih.gov/entrez/query.fcgi?…`, which otherwise counts as prose) ÷ total.

| rule | fires when | catches |
|---|---|---|
| `pubmed-density` | ≥ 2 complete PubMed links **and** prose < 60 % | every PDQ reference block, including ones cut mid-entry (production `::25`, `::26`) |
| `references-block` | `References` heading + ≥ 2 numbered entries, prose < 70 % | a block whose links were truncated by the chunk boundary |
| `mid-block-entries` | ≥ 3 numbered reference entries + ≥ 1 PubMed link, prose < 60 % | a chunk that starts inside a truncated `PUBMED Abstract]](…)` marker |

Kept, by test: prose with inline `[[6](#cit/section_3.6)]` markers; a paragraph that ends with one PubMed link; ordinary numbered checklists (the entry pattern requires author-comma-author or a year); Hindi/Hinglish text; empty chunks.

## Where it runs

`rag.service.ts` → `withoutReferenceChunks()` at the point where SQL rows become evidence, **before any ranking** and inside each path's 2× over-fetch, in all four SQL-backed paths — vector, FTS, keyword, cancer-type — so hybrid and multi-query inherit it. Logs `reference_chunks_dropped` (stage, counts, first chunk ids, query prefix). Takes effect on the next deploy; no re-ingest. The ingest-time fix (strip `References` before chunking) stays with the #86 re-index, per review.

## Tests

- Fixtures are cut from the **real** pregnancy PDQ at the ingest chunk size (~1 400 chars): the heading block, a mid-entry block, the chemotherapy prose, the lactation/fetal section.
- **#126 ranking regression**: the production top-4 with real scores → after the filter, `pregnancy::18` (chemotherapy) is first; `::26`/`::25` are the only drops.
- **KB-wide sanity** over a sample of PDQ documents (1 209 prose chunks, 100 reference chunks): **0 prose chunks flagged** (absolute) and ≥ 98 % of reference blocks caught (the residual is a block citing ACS/SEER reports by bare PDF URL).
- `reference-chunk-filter.spec.ts` 13/13; full API suite **61 suites / 1 091 tests**; `tsc` clean; repo guards OK.

## After deploy

Replay the #126 probe on web and WhatsApp: the pregnancy question should now lead with chunk `::18`-grounded fetal-risk content; the breastfeeding control (probe D) must be unchanged. Watch `reference_chunks_dropped` volume in logs for the first day.

### DIFF
--- a/apps/api/src/modules/rag/__fixtures__/pdq-pregnancy-breast-treatment.excerpt.md
+++ b/apps/api/src/modules/rag/__fixtures__/pdq-pregnancy-breast-treatment.excerpt.md
@@ -1,0 +1,75 @@
+<!--
+Regression fixture for apps/api/src/modules/rag/reference-chunk-filter.spec.ts (issue #129 / #126).
+Verbatim excerpts of the NCI PDQ summary "Breast Cancer Treatment During Pregnancy"
+(Health Professional Version), a U.S. federal government work in the public domain
+(https://www.cancer.gov/publications/pdq). Only the three sections the test needs are
+kept; nothing is edited inside a section. The full document lives in the KB corpus
+(kb/en/, not tracked in git), which is why the test cannot read it in CI.
+Sections are separated by the marker lines below; the spec splits on them.
+-->
+
+<!-- FIXTURE:references-block -->
+###### References
+
+1. Hoover HC: Breast cancer during pregnancy and lactation. Surg Clin North Am 70 (5): 1151-63, 1990. [[PUBMED Abstract]](http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids=2218825&dopt=Abstract "http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids=2218825&dopt=Abstract")
+2. Gwyn K, Theriault R: Breast cancer during pregnancy. Oncology (Huntingt) 15 (1): 39-46; discussion 46, 49-51, 2001. [[PUBMED Abstract]](http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids=11271981&dopt=Abstract "http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids=11271981&dopt=Abstract")
+3. Moore HC, Foster RS: Breast cancer and pregnancy. Semin Oncol 27 (6): 646-53, 2000. [[PUBMED Abstract]](http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids=11130471&dopt=Abstract "http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids=11130471&dopt=Abstract")
+4. Rugo HS: Management of breast cancer diagnosed during pregnancy. Curr Treat Options Oncol 4 (2): 165-73, 2003. [[PUBMED Abstract]](http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids=12594943&dopt=Abstract "http://www.ncbi.nlm.nih.gov/entrez/query.fcgi?cmd=Retrieve&db=PubMed&list_uids=12594943&dopt=Abstract")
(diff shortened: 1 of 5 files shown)


gautamgauri:
Both review findings fixed in `7b69755`:

1. **CI ENOENT** — `kb/en` is git-ignored, so the corpus is absent in CI. The three PDQ sections the tests need are now committed verbatim under `src/modules/rag/__fixtures__/pdq-pregnancy-breast-treatment.excerpt.md` (NCI PDQ, U.S. public domain; nothing edited inside a section). The corpus-wide sanity check runs only where `kb/en` exists (`describe.skip` otherwise, with the directory read guarded because Jest executes skipped describe bodies at collection time). Verified: 13/13 with the corpus present, 12 passed + 1 skipped with it hidden.
2. **Query text in logs** — `reference_chunks_dropped` now carries only `stage`, `dropped`, `kept` and the first six `droppedChunkIds`; the `query` parameter is gone from `withoutReferenceChunks`.

RAG suites 63/63, `tsc` clean.

gautamgauri:
Review update on head `7b69755`: **READY**.

Both blockers from the previous pass are fixed cleanly:

- The #126 regression no longer depends on the git-ignored local KB corpus. The committed NCI-PDQ excerpt gives CI a deterministic fixture; the broader corpus sanity check is correctly guarded/skipped when `kb/en` is absent.
- `reference_chunks_dropped` no longer logs any patient query text — only stage/counts/chunk IDs.

I also re-checked the wiring: the filter is applied before trusted-source reranking in the vector/keyword paths and on mapped results in the FTS/cancer-type paths, so the hybrid/multi-query paths inherit it rather than needing a second filter.

Required API CI is green (unit tests + build/config parity). The Tier1 workflow is red, but that lane still evaluates the live deployed service rather than this PR head, so it is not evidence of a branch regression here.

One operational check remains after merge/deploy, as the PR already says: replay #126 on web + WhatsApp and verify the pregnancy answer is grounded in the chemotherapy/fetal-risk prose while the breastfeeding control is unchanged.

PR #133, opened 2026-09-12
fix(rag,db): full-text search as a GIN expression index — no tsvector column, no backfill (#92 redesign, replaces #130's column design)

gautamgauri:
Refs #92, #98, #130 (review comment `5648272849`), #126. **Production is already on this design** (see below); this PR makes the code match it.

## Why the redesign

Both column-based designs failed on production on 2026-09-12:

| design | what happened |
|---|---|
| STORED generated column (#98 as merged) | full table rewrite under ACCESS EXCLUSIVE lock, rebuilding every index incl. the 182 MB pgvector HNSW index → **15-minute retrieval outage**, terminated |
| plain column + trigger + batched backfill (#130) | non-blocking, but each updated row is too large for a HOT update and re-enters all four indexes incl. HNSW → **190 ms/row, ~2.5 h**; stopped at 12k rows, and a half-filled column meant hybrid retrieval saw lexical hits from a subset |

**Expression index:** `GIN (to_tsvector('simple', content))` + a query on the identical expression. No column, no trigger, no backfill, no per-row writes, no shadow state to drift, nothing Prisma has to represent. Without the index the query still returns correct rows (sequential scan), so the lexical arm can only be *dead* if the table is missing.

## What was done on production (with the maintainer's go-ahead, 2026-09-12 ~19:45 UTC)

1. Backfill stopped permanently; trigger, function and plain column dropped — catalog-only, **3 s**, zero lock waiters.
2. `CREATE INDEX CONCURRENTLY kb_chunk_content_tsv_idx ON "KbChunk" USING GIN (to_tsvector('simple', content))` — **90 s**, 36 MB, `indisvalid`/`indisready` true.
3. Verified: definition is `to_tsvector('simple'::regconfig, content)`; live search `cancer treatment` → **11,791** rows; `EXPLAIN` → **Bitmap Index Scan on kb_chunk_content_tsv_idx**.
4. `prisma migrate resolve --applied 20260908000000_restore_kb_chunk_fts`; the rewritten `migration.sql` re-runs as a no-op on the database.

Health stayed `ok` throughout. The running revision (`00419`) still queries `c.content_tsv`, so `/v1/health/retrieval` reports *unavailable* (clean vector-only fallback, no partial lexical results) until this PR deploys.

## Changes

| file | change |
|---|---|
| `kb-fts.sql.ts` | `KB_FTS_EXPRESSION` is the one place the expression lives; `KB_FTS_SEARCH_SQL` uses it in predicate and rank; probe returns `tablePresent`, `indexDef`, `indexValid`, `legacyColumnPresent`; column/trigger constants removed |
| `KbFtsHealthService` | table missing → `unavailable`; index missing / INVALID / wrong expression or config → `degraded`; else `ok`; a lingering legacy column is reported so it gets dropped |
| `migration.sql` (same name — no persistent DB had recorded it) | one atomic DO block: drop leftovers of both abandoned designs → build the expression index on ≤ 5 000 rows → **raise** on a larger table unless a valid expression index already exists |
| `schema.prisma` | `content_tsv` field and its index mapping removed; the guard note stays (a schema diff *will* propose dropping the index — never accept it) |
| `scripts/sql/kb_fts_safe_rollout.py` | drop legacy → `CREATE INDEX CONCURRENTLY` → verify (valid, expression, live hits, `EXPLAIN` uses the index) → `migrate resolve` → no-op proof; dry run by default |
| runbook §4a, backlog incident record | both abandoned designs and the final one |

## Tests (PGlite, real migrations replayed oldest-first)

Index valid and over the right expression · planner uses it (`enable_seqscan = off`) · new rows searchable at once · migration removes a legacy column/trigger/function and builds the index · **no index → rows still returned, probe `degraded`, and a query success does not flip a stale `unavailable` to `ok`** · INVALID index → `degraded` · table gone → `42P01` classified, `unavailable`, escalated · 5 001-row refusal leaves no index · no-op after a valid index exists · schema guard. `kb-fts.spec.ts` **20/20**; RAG + health suites 51/51; `tsc` clean; `prisma validate` ok; repo guards ok.

## After merge

Deploy; `GET /v1/health/retrieval` should read `ok` at boot; then replay the #126 probes (and the Tier-1 retrieval eval) with the lexical arm live.

### DIFF
--- a/apps/api/prisma/migrations/20260908000000_restore_kb_chunk_fts/migration.sql
+++ b/apps/api/prisma/migrations/20260908000000_restore_kb_chunk_fts/migration.sql
@@ -1,4 +1,4 @@
--- Restore full-text search on KbChunk (issue #92) — NON-BLOCKING implementation.
+-- Restore full-text search on KbChunk (issue #92) — EXPRESSION INDEX, no column.
 --
 -- HISTORY
 --   20260120163141_add_fts_to_kbchunk  added `content_tsv` as a STORED generated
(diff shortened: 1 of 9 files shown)


gautamgauri:
Review status: **HOLD — expression-index design is the right direction, but degraded FTS currently falls into a full-table lexical scan on every hybrid turn.**

The production index itself is healthy, and the new query/index expression match is correct. The blocker is the failure mode this PR deliberately introduces for a missing / INVALID / mismatched index:

- `KbFtsHealthService` reports that state as `degraded`, correctly noting that the query can still return rows via sequential scan.
- But `RagService.fullTextSearchWithMetadata()` does **not consult that health state before executing `KB_FTS_SEARCH_SQL`**. Hybrid retrieval calls vector + FTS in parallel on every turn.
- Therefore an index loss/regression no longer gives us the old fast schema error → vector-only fallback. It makes every lexical arm compute `to_tsvector('simple', c.content)` across the full ~49k-row `KbChunk` table and rank matches. On the current `db-f1-micro`, after the two latency incidents we just measured, that is not a safe degraded mode: it can turn an FTS-index problem into per-turn DB CPU/latency/pool pressure.

Please make the runtime fail *operationally safe*: if the boot/re-probe verdict is `degraded` because the expression index is missing, invalid, or mismatched, skip the FTS arm and return `[]` (vector-only) until the probe returns `ok`. `unavailable` should likewise skip. `unknown` can remain a deliberate choice (try once vs skip) but should be explicit. Add a regression proving that a missing index does **not** execute the expensive lexical SQL while health is degraded.

This preserves the major advantage of the expression design — no shadow column and no backfill — without making sequential full-text scans the steady-state fallback.

**Non-blocking robustness while touching the rollout script:** `Db.prisma_url()` still uses `urllib.parse.quote()` with its default `safe='/'`, so a username/password containing `/` reconstructs an invalid Prisma URL. Use `quote(..., safe='')` (and encode the DB-name component) as previously flagged. Production happened not to hit this, but the runbook script should be reusable safely.

Everything else I checked in the redesign is directionally sound: one expression constant in the shipped query, valid/index-expression probe, large-table migration refusal, concurrent production build, and planner-use verification.

gautamgauri:
Both points fixed in `b6d155e`.

**Operational fallback.** `KbFtsHealthService.shouldQuery()` returns true only for an `ok` verdict; `degraded`, `unavailable` **and** `unknown` all return false (skipping on `unknown` is deliberate — one throttled probe beats a blind full scan), log `kb_fts_arm_skipped` (throttled to once a minute) and schedule a throttled re-probe, so the arm resumes by itself once a valid expression index is seen. `RagService.fullTextSearchWithMetadata()` consults the gate before running `KB_FTS_SEARCH_SQL` and answers vector-only (`[]`) otherwise. The steady-state fallback is therefore vector-only, never a per-turn sequential `to_tsvector()` scan.

**Regression tests (PGlite):** with the index missing the lexical SQL is *not executed* (spy on `$queryRawUnsafe` shows zero `websearch_to_tsquery` calls) and the verdict stays `degraded`; after `CREATE INDEX` and a re-probe, `shouldQuery()` flips and the arm returns rows; a stale `unavailable` verdict on a healthy schema is cleared by the gate's re-probe and the next call runs; a table dropped *after* a healthy boot still classifies `42P01` and escalates. 21/21.

**Rollout script URL encoding:** `prisma_url()` now quotes user, password and dbname with `safe=''`, and the dbname is unquoted when the URL is parsed, so reserved characters round-trip (`us/er` → `us%2Fer`, `db/name` → `db%2Fname`).

PR #139, opened 2026-09-13
fix(chat): empty-bold cleanup must not cross a line break — escalation header kept its paragraph break (#135)

### LINKED ISSUE #135: Escalated reply (bubble) loses the header line break: display cleanup eats "**\n\n**" so the emergency header renders as "medical emergency.Call for help NOW:"
## Summary

On the **critical** escalation path the assistant reply's header loses its paragraph break inside the message bubble: the patient reads

> ⚠️ **This sounds like a medical emergency.Call for help NOW:**

with `emergency.` and `Call` glued into a single bold run, instead of the intended `…medical emergency.` / *blank line* / `Call for help NOW:`.

The emergency **banner** is correct — this is the bubble only. It is cosmetic (no clinical copy changes, no missing escalation step), but it is on the highest-stakes path and makes the reply look broken exactly when a caregiver is panicking.

## Evidence — live prod, 2026-09-13 (browser QA, Playwright/Chromium)

- seed `1789272008`, RUN_DIR `~/.hermes/suchi-daily/runs/2026-09-13T04-00-12_seed1789272008/`
- q02 `messageId c8f0e493-87c1-439d-b359-7b059e4e50fb` — *"My mother has cancer and suddenly she can't breathe properly and her face is swollen. What should I do right now?"*
- `safety: {"classification":"red_flag","actions":["show_emergency_banner","end_conversation"]}`
- In one `/chat` payload, the two text fields disagree:

| field | len | first 70 chars |
|---|---|---|
| `responseText` (bubble) | 919 | `⚠️ **This sounds like a medical emergency.Call for help NOW:**\n• **112** …` |
| `safety.bannerText` (banner) | 925 | `⚠️ **This sounds like a medical emergency.**\n\n**Call for help NOW:**\n• **112** …` |

The delta is exactly 6 characters: `**\n\n**` → ``. DOM read of the bubble in the same session returns the glued text (`This sounds like a medical emergency.Call for help NOW:`), so it renders that way for the patient, not just in the payload.

## Mechanism (in-repo, verified against the code)

`apps/api/src/modules/chat/chat.controller.ts:34` runs on `result.responseText` only (the banner gets `safety.bannerText`, which is never passed through it):

```ts
// Clean up any leftover empty bold markers like "** **"
.replace(/\*\*\s*\*\*/g, "")
```

The `\s*` also matches newlines, so it eats the boundary between *two legitimate adjacent bold spans* (`…emergency.**` + `\n\n` + `**Call for help NOW:**`) — not just a truly empty `** **`. The critical template at `apps/api/src/modules/safety/emergency-fast-path.ts:119-121` is the only escalation copy with that `**…**\n\n**…**` shape; the S2/urgent template (`response-templates.ts:297-299`) has no closing `**` before its break, which is why this never showed up in the earlier escalated QA samples (2026-09-10/09-12 all began `Some of what you described could be urgent.`).

Not verified: the deployed revision is far ahead of this checkout (`HEAD` is `8d15e3f`; prod was `suchi-api-00420-9sr`), so the line number is from the local tree — the transformation itself is reproduced in the payload.

## Suggested constraint (code fix, no clinical wording touched)

Narrow the empty-marker cleanup so it cannot cross a line, e.g. `/\*\*[ \t]*\*\*/g` (and/or require the span to be genuinely empty). Escalation copy itself needs no change.

## Related

- #111 (closed) — the banner half of this path; the banner is now correct, this is the remaining bubble-side artifact. Note it also breaks #111's documented invariant that `bannerText` is a byte-exact prefix of `responseText`.
- #112 — separate (unreconciled appended RAG half); not involved here, nothing was appended on this turn (0 citations).

gautamgauri:
Fixes #135.

## What the payload showed

Run `2026-09-13T04-00-12_seed1789272008`, `q02 [red_flag]`, messageId `c8f0e493-…`. In one `/chat` payload the raw escalation text (`safety.bannerText`, 925 chars, never cleaned) and the bubble text (`responseText`, 919 chars) differ by exactly the six characters `**\n\n**` between the header line and the "Call for help NOW:" line.

## Cause

`apps/api/src/common/text-cleaning.ts`, applied by `cleanResponseForDisplay` (`display-text-cleaner.ts`) from `ChatController.send()`:

```ts
const EMPTY_BOLD_PATTERN = /\*\*\s*\*\*/g;
```

`\s*` matches newlines, so `emergency.**` + `\n\n` + `**Call` — two legitimate adjacent bold spans — is matched as one "empty bold" and deleted. The critical template in `emergency-fast-path.ts` is the only escalation copy with the `**…**\n\n**…**` shape, which is why earlier escalated samples (S2/urgent template, no closing `**` before its break) never showed it.

Note: the issue quotes `chat.controller.ts:34` from an older checkout; on `origin/main` the pattern had already moved to the shared `common/text-cleaning.ts` (PR for #87), so this fix lands there and covers every consumer (display cleaner, voice stripper, condenser, copilot patch executor).

## Fix

```ts
const EMPTY_BOLD_PATTERN = /\*\*[ \t]*\*\*/g;
```

Horizontal whitespace only. A genuinely empty `** **` / `****` on one line is still removed (tested). No template, escalation or clinical wording changed.

## Tests (exact strings from the run; all red before the fix, green after)

- `common/text-cleaning.spec.ts` — `stripCitationDebris`: keeps `…emergency.**\n\n**Call for help NOW:**` intact; also single `\n`; still strips one-line empty bold.
- `chat/display-text-cleaner.spec.ts` — the full 925-char `bannerText` from the run passes through **byte-identical**; header and "Call for help NOW:" stay on separate lines.
- `chat/chat.controller.spec.ts` — at the controller boundary, `responseText.startsWith(safety.bannerText)` holds again (the #111 invariant) for a red_flag result.

```
cd apps/api && npx jest --testPathPatterns="text-cleaning|display-text-cleaner|chat.controller.spec|voice-output-stripper|voice-response-condenser" -w 2
# before fix: 5 failed, 48 passed (the 5 new tests)   after fix: 5 suites, 83 passed
cd apps/api && npx jest -w 2   # full suite: see final summary
```

## The stray leading `S` — separate, and not a rendering defect

**Evidence:** every one of the six replies in the run (`q01`–`q06`, escalated or not) has `reply_dom` starting with `S`. The harness reads `page.locator('[aria-label="assistant message"]').textContent()` (`runner.mjs:111,145`). That element is the outer message row in `apps/web/src/components/MessageList.tsx`, which contains `<SuchiAvatar size="small" />` *before* the bubble, and `SuchiAvatar.tsx` renders the literal letter `S` as the avatar glyph (`role="img"`, `aria-label="Suchi avatar"`). `textContent` concatenates the avatar's `S` with the bubble text. So the `S` is the avatar letter picked up by the harness's DOM read, present on all replies, unrelated to the `**\n\n**` strip. Not changed here (frontend, out of this issue's scope); if the harness wants a clean read it should target the bubble node, or the avatar glyph could be `aria-hidden` — either is a separate ticket.

## Not touched

- `voice.service.ts` no longer carries its own copy of this pattern on `origin/main` (it calls the shared cleaner), so no second fix was needed.
- #112 (appended RAG half) — nothing appended on this turn (0 citations).


gautamgauri:
Review status: **READY.**

This is a narrow, correct fix for #135. Restricting the empty-bold matcher from `\s*` to horizontal whitespace `[ \t]*` preserves legitimate paragraph/line breaks between adjacent bold spans while still removing `****` / `** **` debris on one line.

The regression coverage is appropriate at three levels: shared cleaner, display cleaner using the exact production banner shape, and controller boundary preserving the #111 `bannerText` prefix invariant. The separate leading `S` from the avatar/harness DOM read is correctly diagnosed as out of scope rather than mixed into this fix.

Required API CI is green. I found no merge blocker.

Issue #140, opened 2026-09-13
Landing /about: founders_photo.jpg is a Facebook HTML page saved as .jpg — the founders photo is broken on the live site (logo.jpg is the same, unreferenced)


gautamgauri:
## What happened

`apps/landing/public/assets/founders_photo.jpg` is **not an image**. It is a 152,390-byte Facebook
login/interstitial HTML page saved under a `.jpg` filename, so the founders photo on the public About page
renders as a broken image (only the alt text shows).

```
apps/landing/public/assets/founders_photo.jpg   152390 bytes
  file: HTML document, Unicode text, UTF-8 text, with very long lines (38188)
  head: <!DOCTYPE html>
        <html lang="nl" id="facebook">
```

Same defect in `apps/landing/public/assets/logo.jpg` (152,385 bytes, identical HTML) — though **nothing
references it**, so it is currently dead weight in the deployed bundle rather than a visible breakage.

## Where it is referenced

```
apps/landing/src/pages/about.astro:128
  <img src="/assets/founders_photo.jpg"
       alt="SCCF Founders: Gautam Gauri, Dr. Amit Kumar, and Dr. Nandini Thakur Jha" loading="lazy" />
```

## Verified against the live site (2026-09-13)

```
GET https://suchitracancercare.org/about           -> page contains "founders_photo.jpg"
GET https://suchitracancercare.org/assets/founders_photo.jpg
    HTTP/2 200
    content-type: image/jpeg          <-- claims to be a JPEG
    content-length: 152390
    body first bytes: <!DOCTYPE html>\n<html lang="nl" id="facebook"> ...
```

The server serves it as `image/jpeg` (by extension) while delivering HTML, so every browser fails to
decode it. This is live right now, on the organisation's public website, on the page that introduces the
founders.

`assets/logo` is referenced **0** times on `/` and `/about`, confirming `logo.jpg` is unused.

## Likely cause

A fetch of an image from a Facebook CDN URL that returned a login/interstitial page (note `lang="nl"` and
`id="facebook"`) and was written to disk as if it were the image body — the same failure shape for both
files (sizes differ by 5 bytes) suggests one scripted download run.

## Fix

1. Replace `founders_photo.jpg` with the real photograph. **The original asset is needed from SCCF** — the
   current file cannot be repaired by renaming. Consider resizing/optimising it for the web while you are
   there (it is a hero-style image on a public page).
2. Delete `logo.jpg` unless something is about to use it.
[...]

PR #141, opened 2026-09-13
fix(landing): remove HTML-as-.jpg founders photo/logo, add image magic-byte CI guard (part of #140)

gautamgauri:
## Summary

Part of #140 — the engineering half only. **This PR does NOT fix the founders photo. The real photograph is still needed from SCCF.** It removes the two broken files so the public About page no longer shows a broken image, and adds a guard so an HTML-page-saved-as-.jpg can never ship again.

### What was wrong
`apps/landing/public/assets/founders_photo.jpg` (152,390 B) and `logo.jpg` (152,385 B) were Facebook login/interstitial HTML pages (`<html lang="nl" id="facebook">`) saved under a `.jpg` name. GitHub Pages served them as `image/jpeg`, browsers failed to decode, and the founders photo on https://suchitracancercare.org/about rendered as alt text only. `logo.jpg` was referenced nowhere (verified with grep).

### Changes
- **Delete** `apps/landing/public/assets/logo.jpg` (unreferenced) and `founders_photo.jpg` (broken).
- **`apps/landing/src/pages/about.astro`**: remove the ``; leave an Astro `{/* */}` comment (stripped at build) explaining the asset is pending from SCCF (issue #140) and showing the exact markup to restore. The founders' names and credits were already visible as `<h3>` text in the founder cards, so nothing user-visible is lost. `.founders-photo` CSS kept for the restore.
- **New `scripts/check_image_assets.py`**: every git-tracked image-named file under `apps/landing/public/` and `apps/web/public/` must have real magic bytes — JPEG `FF D8 FF`, PNG `89 50 4E 47 0D 0A 1A 0A`, GIF87a/89a, `RIFF....WEBP`, ICO (`00 00 01 00` or PNG-in-ICO, which Astro's default favicon is), AVIF, BMP; SVG must start with `<svg` or `<?xml` and contain no `<html`/`<head`/`<body`/`<script`. Any image-named file opening with `<!DOCTYPE html>`/`<html` fails regardless of extension. Exits non-zero listing offenders. `--self-test` runs 12 good / 12 bad byte heads plus an end-to-end HTML-as-.jpg case in a throwaway git repo.
- **CI**: added to `.github/workflows/deploy-api.yml` "Build + config parity" job (the required PR gate, right after `check_deploy_config_parity.py`) and to the `deploy-landing.yml` build job before `npm ci`, so a bad asset can neither merge nor deploy. `cloudbuild.yaml` / `cloudbuild.gated.yaml` untouched.

### Verification
```
$ cd apps/landing && npm ci && npm run build
[build] 84 page(s) built in 1.34s — Complete!
# dist/about/index.html: 0 <img> tags, 0 "founders_photo" mentions, all three founder names present

$ python3 scripts/check_image_assets.py
check_image_assets: 3 tracked image file(s) under apps/landing/public, apps/web/public
OK: every tracked image-named file has valid image magic bytes

$ python3 scripts/check_image_assets.py --self-test
self-test: OK (12 good heads accepted, 12 bad heads rejected, end-to-end HTML-as-.jpg detected)

# demo: HTML saved as .jpg, git-added, then removed
FAIL: 1 file(s) are not the image their extension claims:
  apps/landing/public/assets/demo_fake.jpg  (85 bytes): HTML document saved with an image extension
exit=1
```

### Still open (human / SCCF)
- [ ] SCCF supplies the real founders photograph (resize/optimise for web) — then add it to `apps/landing/public/assets/` and restore the markup in the `about.astro` comment. The guard will verify it is a real JPEG.


gautamgauri (commented):
**HOLD — small CI-guard correctness fix before merge.** The asset cleanup itself is good and removes the broken HTML-as-JPEG files. The remaining blocker is the new required image guard: it rejects valid SVGs that start with a legal XML comment/DOCTYPE before `<svg>`, despite claiming to support normal SVG preambles. Because this check is now on required CI/deploy paths, that false-negative becomes repository-wide friction. Please accept permitted XML comments/declarations/DOCTYPE before the root element and add self-tests for those shapes. After that, this looks ready.

gautamgauri:
Fixed in 1a56a51. The guard required an SVG head to start with `<svg` or `<?xml`, so a comment-first (`<!-- Generator: ... -->`) or DOCTYPE-first SVG was rejected — the false positive you flagged, and contrary to the `HEAD_BYTES` comment that already claimed those preambles were supported.

New `svg_root_element()` steps over a BOM, whitespace, XML declarations and processing instructions, XML comments, and a DOCTYPE (including one with an internal `[...]` subset), then returns the name of whatever element actually comes first; the SVG branch requires that root to be `svg`. HTML markup anywhere in the head (`<html`, `<head`, `<body`, `<script`, `<!doctype html`) is still rejected first, so a comment-laundered HTML page (`<!-- innocent --><html>...`) does not sneak through. A prolog that never terminates within `HEAD_BYTES` fails rather than being assumed good.

Self-test corpus 12/12 → 18 good / 17 bad. Added: comment-first, multi-line comments, DOCTYPE-first, DOCTYPE-with-internal-subset, BOM + declaration + comment + DOCTYPE, and an `xml-stylesheet` PI (all accepted); comment-then-HTML, comment-then-non-svg-root, non-svg DOCTYPE, and an unterminated comment (all rejected).

`python3 scripts/check_image_assets.py` → OK, 3 tracked image files. `--self-test` → OK, exit 0.

gautamgauri:
Re-review on head `1a56a51`: **READY**. The SVG false-positive blocker is closed: `svg_root_element()` now walks BOM/whitespace, XML declarations/PIs, comments and DOCTYPE before requiring an `<svg>` root, while HTML-shaped content is still rejected. The expanded self-test covers the previously missed preambles. Required `API unit tests` and `Build + config parity` are both green on this head.

gautamgauri (commented):
One concrete gap in the new asset guard: for raster formats it validates only the leading signature. A truncated file containing just a valid PNG/JPEG/GIF/etc. header will therefore pass even though no browser can decode it. The original HTML-as-.jpg failure is caught, but the PR currently describes the guard more broadly as ensuring tracked image assets are real/valid images. Please either decode/verify the complete raster (for example with Pillow or another lightweight verifier) or narrow the guarantee/documentation and add a truncated-file regression case.

PR #142, opened 2026-09-13
fix(eval): judge transport failures are unscored, not failed rubric checks (#110)

### LINKED ISSUE #110: Tier1 eval: judge transport failures (429/ClientError) are scored as failed checks — failing closed on noise
Filed from the open-wave delegation findings (2026-09-09, after #74 revived the judge).

## What happens
Now that the Tier1 judge actually runs (post-#74), every PR touching rag/evidence shows 'Run Tier1 Retrieval Quality Eval = FAILURE' — but the failures are NOT branch regressions. Two independent causes:

1. **Transport noise is scored as failed checks.** Vertex AI `429 RESOURCE_EXHAUSTED` responses are recorded as failed judge checks. Concrete case: RQ-PROSTATE-02 had ALL SIX judge checks return 429 and landed at 36.4%. Some verdicts arrive as `unknown:` / ClientError shapes that are likewise scored as case failures.
2. **Genuine prod coverage misses (~91% avg)** — real content_coverage misses that deserve their own attention (separate from this infra issue).

## Evidence it is not branch-specific
`eval-tier1.yml:81` points EVAL_API_BASE_URL at the LIVE Cloud Run service, so on a PR the check evaluates production, never the branch. Control case: PR #58 (line-endings-only change) scores the identical 6/21. Counts also drift 6/6/5/6 across identical targets. The 3->6 shift vs main's pre-#74 nightly is the judge waking up — revealing, not causing.

## Recommended fix (for the delegation/SCCF-neutral code path)
Classify `unknown:` / ClientError judge outcomes as *errored* (not failed checks), retry 429s, and route persistent transport failures to the workflow's existing 'infrastructure error' state — the fail-closed inverse of #78's fail-open bug. Note the PR gate's deeper oddity: Tier1 evaluates prod, never the branch — so as a PR gate it measures drift, not the PR; worth a separate decision on what the PR check should gate.

gautamgauri:
## Problem

The Tier1 LLM judge fails **closed** on infrastructure noise.

A Vertex `ClientError` looks like this — note there is no `status` property; the HTTP code lives in `cause.code` and in the message text:

```
[VertexAI.ClientError]: got status: 429 Too Many Requests. {"error":{...}}
```

The old classifier in `llm-judge.ts` only inspected `error.status` / `error.response.status` and the literal phrase `"rate limit"`, so a Vertex 429 fell through to the `unknown` branch — the one branch that deliberately does **not** skip — and every check on the case was written out as a **failed rubric check**. RQ-PROSTATE-02 had all six checks come back 429 and was scored 36.4%: a quality failure manufactured from quota noise. The nightly and the PR gate then went red for a retrieval regression that never happened.

## Change

A judge call now ends one of exactly two ways: it renders a verdict, or it does not. Only the first says anything about answer quality.

**`eval/runner/judge-errors.ts` (new)**
- `classifyJudgeError()` — handles Vertex (`cause.code` + message), OpenAI (`status`), axios (`response.status`), gRPC (`RESOURCE_EXHAUSTED`, `DEADLINE_EXCEEDED`, `UNAVAILABLE`, …) and Node (`ECONNRESET`, `ETIMEDOUT`, …) shapes → `rate_limited | timeout | provider_error | network_error | auth_failed | sdk_missing | not_configured | malformed_verdict | unknown`, each tagged `transport` and `retryable`. Never throws.
- `callJudgeWithRetry()` — **bounded retry with exponential backoff + jitter** on retryable errors, honouring a server `Retry-After` header or Vertex `retryDelay` hint, capped at 30s. `EVAL_JUDGE_RETRIES` (default 3 extra attempts). Auth failures and unknown errors are not retried.
- `unscoredJudgeResult()` / `resolveCaseUnscored()` — the `unscored` result shape and the case-level bucket decision.

**`llm-judge.ts`** — retries, then returns UNSCORED results with `unscoredReason` + `attempts` instead of failed checks. Also: a same-provider "fallback" no longer runs as an extra un-backed-off attempt; an unparseable or incomplete reply is a *missing* verdict, not a fail verdict; consensus no longer throws away a real verdict because the other run went unscored.

**`report-generator.ts`** — three mutually exclusive buckets (`passed` / `failed` / `unscored`) plus a `summary.judge` availability block: `status` (`active|degraded|unavailable|not_run`), scored/unscored check counts, affected case ids, and reason histogram. Unscored checks stay out of the score's numerator **and** denominator, and a case with a missing verdict can no longer be declared a pass — it is not counted as a failure either.

**Precedence — rendered evidence outranks a missing verdict.** A case that errored out, missed a *required* deterministic check, or failed a *required* judge check that really did answer `ok: false` stays a genuine **failure** even when a sibling check went unscored. A judge hole cannot launder a real quality failure into "judge unavailable". The evaluator decides this where it holds the rubric and records the decision explicitly on the result; the report and records read that decision.

**Per-case records / clusters** — `outcome.unscored`, `outcome.unscoredReason`, `outcome.unscoredLlmChecks`, and a `judge-unavailable` failure cluster, counted separately from `failedCases`.

**`failure-miner.ts`** — unscored cases are not mined for repairs; there is nothing to patch when the judge never answered.

**`ci/eval-status.js` — the distinct CI signal.** New `judge_unavailable` status:

| situation | status | CI |
|---|---|---|
| unscored cases **above** `EVAL_MAX_UNSCORED_RATIO` (default `0.1`) | `judge_unavailable` | **exit 1**, `::error title=Judge unavailable::… infrastructure failure, not a quality failure` |
| unscored cases within tolerance | `passed` / `failed` as usual | `::warning title=Judge partially unavailable::` — green, and those cases are *unscored*, not passed |
| real case failures on rendered checks | `failed` | exit 1, `::error title=Evaluation failed::` (unchanged) |
| no report | `infra_error` | exit 1 (unchanged) |

Two numbers are tracked deliberately and not merged: the **outcome bucket** (`evaluation.unscored` / `unscoredCaseIds`, never counted as failures) and the **availability axis** (`evaluation.judge.unscoredCases` — every case the judge left a hole in, including ones that also failed on checks it *did* render). The tolerance gates on the availability axis, so a judge outage is still called an outage; `failedCaseIds` excludes only the outcome bucket. `judge_unavailable` outranks `failed` on purpose — when the judge could not score enough of the run, the red should say so rather than point at whatever subset happened to fail (the annotation still names the genuine failures it observed).

The job summary and the notification email list unscored cases in their own section, with judge status, reasons and tolerance. `merge-reports.js` gets a matching `unscored` bucket.

**`eval-tier1.yml`** — documents the fourth state, sets `EVAL_MAX_UNSCORED_RATIO=0.1` and `EVAL_JUDGE_RETRIES=3`, and passes the tolerance to the `result` subcommand. Status handling still reads **only** `eval-result.json`; the `check` subcommand remains the single source of truth for the job conclusion.

**Not changed:** rubric weights, thresholds, check definitions, `eval/cases/**`, `eval/rubrics/**`, and every safety/deterministic check. No file under `apps/api/` is touched.

## Test evidence

```
$ cd eval && npx tsc --noEmit
(clean)

$ npx jest
Test Suites: 10 passed, 10 total
Tests:       141 passed, 141 total

$ node --test ci/          # npm run test:ci
# tests 28
# pass 28
# fail 0
```

New tests:

- **`eval/runner/judge-errors.test.ts`** — classification: the exact Vertex `ClientError` 429 shape (the RQ-PROSTATE-02 regression, asserted as `rate_limited`/retryable/transport, *not* a verdict), status extraction across SDK shapes, `Retry-After` and `retryDelay` hints, 5xx/UNAVAILABLE, timeouts, network errors, auth (transport but **not** retryable), missing SDK, unknown fallthrough, and never throwing on `null`/strings. Backoff growth, jitter bound, cap, and server-hint honouring. `callJudgeWithRetry` retry-then-succeed, give-up-after-N, no-retry-on-auth, `maxRetries: 0`.
- **`eval/runner/judge-aggregation.test.ts`** — scoring (a rate-limited judge no longer drags a clean answer to 36%, while a genuine fail verdict still scores 0.6), pass/unscored determination, all four **precedence** cases, report buckets + judge summary (`active`/`degraded`/`unavailable`), `failures[]` excluding unscored, summary text wording, per-case records and the `judge-unavailable` cluster, and `LLMJudge.judge()` end-to-end: persistent 429 → 3 attempts → unscored checks with `attempts: 3`; retry-then-succeed → real verdicts; unconfigured judge → `not_configured`; unparseable reply → `malformed_verdict`; genuine `ok: false` → still a failure.
- **`eval/ci/eval-status.test.js`** (+13 cases) — tolerance boundary both ways, `judge_unavailable` never reported as `failed`, configurable ratio via flag and env (garbage falls back to the default rather than disabling the gate), unscored ids excluded from `failedCaseIds`, legacy pre-#110 reports still deriving a status, `evaluateCheck` annotations for each state, job-summary and email content.

## Open questions (not addressed here — deliberately out of scope)

1. **The PR gate still evaluates production, not the branch.** `eval-tier1.yml` points `EVAL_API_BASE_URL` at the live Cloud Run service, so on a PR this check measures prod drift, never the PR's code (issue #110 notes this; PR #58, a line-endings-only change, scored identically). Worth its own decision about what the PR check should gate.
2. **The ~91% content-coverage misses are real** and remain — this PR only stops infrastructure noise from being counted alongside them.
3. **Tolerance value.** `0.1` = 2 of the 21 Tier1 cases. Chosen so a single flaky case does not redden the nightly while a quota wall still does; easy to retune via `EVAL_MAX_UNSCORED_RATIO` without a code change.
4. **`determinePass` no longer passes a case with a missing verdict.** Previously a skipped check was excluded and the case could pass on the remaining checks — the fail-open inflation #74 measured. Making "unscored" a real bucket requires this, but it means a run with a fully dead judge now reports N unscored instead of N passed. That is the honest reading, and it is what makes the `judge_unavailable` signal fire.

Fixes #110


gautamgauri (commented):
**HOLD — the direction is right, but the current head still has a core false-failure path.** A judge check object with missing/invalid `ok` can still be coerced to `passed: false`, recreating the exact problem this PR is meant to solve: incomplete judge output becomes a quality failure. Treat invalid/missing `ok` as malformed/unscored. Also bound/normalize `EVAL_JUDGE_RETRIES` so `NaN` cannot create an endless retry loop, and map numeric gRPC status codes (8/4/14 etc.) into the advertised retry classifications. Once those are covered by tests, this should be re-reviewed.

gautamgauri:
All three Codex review findings addressed (commits `69a9468`, `627eb8a`, `679c0b4` — head `679c0b4`):

- **P1 `eval/runner/llm-judge.ts` — invalid `ok` treated as a fail.** `parseOkVerdict` now decides what counts as a rendered verdict: booleans, plus the `"true"`/`"false"` strings Vertex/Gemini intermittently returns. A missing, `null`, numeric, free-text or non-object `ok` is `malformed_verdict` → **unscored**, with the offending value named in the reason, instead of boolean-converting to `passed: false`. A rendered `ok: false` still outranks a sibling unscored check in `resolveCaseUnscored`.
- **P2 `eval/config/loader.ts` — NaN retries could loop forever.** `normalizeJudgeRetries()` coerces any env/JSON/caller value to a finite integer in `[0, 10]`, falling back to the documented default (3) for NaN, ±Infinity, booleans, objects and blank strings. `callJudgeWithRetry` normalises `opts.maxRetries` itself rather than `Math.max(0, …)`, so no config path can reintroduce an unbounded loop.
- **P2 `eval/runner/judge-errors.ts` — numeric gRPC codes ended up `unknown`.** `extractGrpcStatus()` resolves a canonical gRPC status name from the named form (`cause.status`) or the numeric one (`code` / `cause.code` / stringified), so 8, 4, 14, 13, 16 and 7 classify as rate_limited / timeout / provider_error / provider_error / auth_failed / auth_failed and follow the retry policy. gRPC (1..16) and HTTP (100..599) cannot collide, so Vertex's `cause.code = 429` still reads as HTTP and errno strings are untouched; the classification carries `grpcStatus` and only falls back to a `(gRPC NAME)` label when no HTTP status was recovered.

Tests: jest **175 passed / 11 suites** (was 156), `npm run test:ci` **28 passed**, `tsc --noEmit` clean. New coverage: `parseOkVerdict` table + end-to-end `LLMJudge.judge` for every bad/good `ok` shape, `normalizeJudgeRetries` + bounded-attempt assertions for NaN/Infinity/1e6/-5, a `loadConfig` test over junk/sane/negative/absurd `EVAL_JUDGE_RETRIES`, and one case per flagged gRPC code plus negative cases (HTTP label wins, code 0 and 99 stay `unknown`).

gautamgauri:
Re-review on head `679c0b4`: **READY**. All three prior blockers are closed: malformed/missing `ok` is now `malformed_verdict` → unscored rather than false; retry counts are normalised/bounded so NaN cannot create an unbounded loop; and numeric gRPC status codes map into the advertised retry classifications. Required `API unit tests` and `Build + config parity` are green. The red Tier1 job remains the live-production lane rather than a branch-code gate, so I am not treating it as a blocker here.

PR #143, opened 2026-09-13
fix(rag): build the lexical tsquery from content words, not every token (#134)

### LINKED ISSUE #134: Retrieval: lexical arm builds an AND-of-every-token tsquery ('simple' config, no stopwords), so natural-language and all Hinglish queries get zero FTS hits
## What

The lexical arm is live again on production (`suchi-api-00420-9sr`, `/v1/health/retrieval` → `ok`, #133) and it does contribute for keyword-style English queries — `HPV test cervical cancer screening` scored `lexSim` up to 1.0 on five of the top-6; 7 of the first 16 hybrid turns had `ftsCount > 0`. But for natural-language questions, and for **every** Hinglish question, it returns nothing, because `KB_FTS_SEARCH_SQL` feeds the raw query to `websearch_to_tsquery('simple', $1)`:

- `websearch_to_tsquery` joins every token with **AND**;
- the `'simple'` config (chosen deliberately so Hindi/Hinglish tokenise) removes **no stopwords** and does **no stemming**.

So a chunk must contain *every* word of the question, including `of`, `the`, `hai`, `kya`, `meri`.

Measured on production (2026-09-12 21:35 UTC):

| query (as sent to FTS) | AND hits (shipped) | OR hits (same words) |
|---|---|---|
| probe A translated: `meri mausi ko cancer hai … medicine se baby affected hoga? exact tell me` | **0** | 30,001 |
| scenario query `cancer treatment during pregnancy effects on the unborn baby fetus` | **0** | 45,477 |
| `early warning signs of breast cancer` | **0** | 30,926 |
| `chemotherapy during pregnancy fetus` | 7 | 10,185 |
| probe D `kya chemotherapy ke dauraan breastfeeding karna safe hai?` | **0** | 7,963 |

Every lexical score in the #126 replays on `00420` is therefore 0 — the arm is healthy and silent. This is pre-existing (the same SQL shipped before June 2026); it is visible now only because the arm works.

## Fix shape (code only, RAG module)

Build the tsquery from **content words**, not the raw sentence:

1. Tokenise the (translated) query; drop English stopwords and the Hinglish/Hindi function-word set the cross-lingual service already knows (`hai, kya, ko, ki, ke, se, meri, mere, aur, wo, karna, …`); keep medical/content tokens (the `MEDICAL_TERM` list in `cross-lingual.service.ts` is a start).
2. Emit `to_tsquery('simple', 'term1 | term2 | …')` (or `websearch_to_tsquery` with explicit ` or `), keeping any quoted phrase as a phrase, and rank with `ts_rank_cd` so chunks matching more terms rank higher. Optionally require ≥ 2 matched terms via `ts_rank_cd` threshold or a `numnode`-style guard to avoid "cancer"-only floods (an OR query over `cancer` alone matches ~45k chunks — the top-K limit and rank handle it, but a minimum-terms floor is cheap).
3. Keep the expression index and the `'simple'` config unchanged — the index still serves the OR query via GIN.
4. Regression: PGlite test asserting the Hinglish probe (`… kya cancer ki medicine se baby affected hoga …`) matches the pregnancy chunk lexically; unit test for the token filter. Re-measure `ftsCount > 0` share in `hybrid_retrieval` logs before/after.

## Why it matters

Hybrid weights give the lexical arm 45 % of the score on long queries and 20 % on short ones (see P2-8 in `docs/RELIABILITY_BACKLOG.md`: weights are not renormalised when one arm returns nothing, so borderline evidence is pushed under the gate). Today that 45 % is a constant zero for every Hinglish user — the primary audience.

## Related
- #92 / #98 / #130 / #133 (lexical arm restored), #126 (replays show `lex 0` on all pregnancy probes), #132 (cross-lingual translation feeds this query), backlog P2-8 (weight renormalisation).

gautamgauri:
## What changed

The lexical arm of hybrid retrieval no longer feeds the raw question to
`websearch_to_tsquery('simple', $1)`. A new module, `apps/api/src/modules/rag/kb-fts-query.ts`,
builds the tsquery from the question's **content words**:

1. **Token filter.** Tokenises the (already cross-lingually translated) query, drops English
   stopwords, the Hinglish/Hindi function-word set (`hai, kya, ko, ki, ke, se, meri, mere, aur,
   wo, karna, …`), kinship words (`mausi`, `bhai`, …) and chat filler. A word that
   `CrossLingualService`'s `MEDICAL_TERM` recognises is **never** dropped, whatever the stopword
   lists say — `MEDICAL_TERM` and the Hinglish marker list are now exported from
   `cross-lingual.service.ts` for that reuse, and the content-bearing Hinglish markers
   (`ilaaj`, `dawai`, `gaanth`, `bukhar`, `lakshan`, `janch`, `kharcha`, `sarkari`) are pinned
   by test as words the filter must never eat.
2. **OR tsquery with a minimum-matched-terms floor.** Emits an OR of two-term conjunctions —
   `('early' & 'warning') | ('early' & 'signs') | ('early' & 'cancer') | …` — so a chunk is a
   candidate only if it holds **two** of the query's terms.
   **Invariant: every conjunction carries at least one non-generic term.** Words that appear in
   nearly every chunk of a cancer KB (`cancer`, `treatment`, `tumour`, `patient`, …) are marked
   generic: they still contribute when paired with a specific term (so `breast cancer` and
   `lung cancer treatment` keep their recall), but two generic words can never form a hit on
   their own. That is the "single-token `cancer` flood" guard.
3. **Safe escaping.** Each term is passed to `to_tsquery` as a *quoted* string, so Postgres
   applies the same parser to the query as `to_tsvector` applied to the content. Single quotes
   and backslashes are escaped; quoted phrases are preserved as `<->` phrase operands; a word
   must contain a letter or a decimal digit to be emitted. Nothing from user text can reach the
   tsquery grammar.
4. **Ranking** stays `ts_rank_cd` over the same expression.
5. **Health probe.** `KbFtsHealthService` now runs the shipped statement with a
   builder-produced tsquery (LIMIT 1) once the catalog checks pass, so `/v1/health/retrieval`
   reflects the real query path — a tsquery-grammar or renamed-function regression becomes a
   probe verdict instead of `ftsCount: 0` in the logs.
6. Nothing else: **no migration, no tsvector column**, the GIN expression index over
   `to_tsvector('simple', content)` and the `'simple'` config are unchanged — only the TEXT
   that goes into `to_tsquery`. Safety module, prompts and KB text untouched.

## Before / after

Measured in the PGlite regression suite against the real statement and the real migrations
(`kb-fts.spec.ts`), with the pre-#134 statement kept side by side as `LEGACY_AND_SQL`:

| query (as sent to FTS) | before (`websearch_to_tsquery`, AND) | after |
|---|---|---|
| translated probe A — `meri mausi ko cancer hai … kya cancer ki medicine se baby affected hoga? exact tell me` | **0 rows** | matches `chunk-pregnancy`, **ranked first** |
| `What should happen after an abnormal screening result at the district hospital?` | **0 rows** | `chunk-followup` first |
| `colposcopy screening follow-up` | — | only `chunk-followup` (a one-word overlap is not a hit) |
| `"unborn baby" chemotherapy` | — | phrase-matched; `"baby unborn"` (wrong order) returns nothing |

Two design choices were settled empirically against a real Postgres rather than by assertion:

- **Pairwise vs plain OR ranking.** `ts_rank_cd` on a plain `a | b | c` ties a chunk matching 4
  terms with one matching 3 (0.3 / 0.3 / 0.2 / 0.2). The pairwise form separates them
  (0.070 / 0.042 / 0.025 / 0.020). Pairwise ranks strictly better, so the floor costs nothing.
- **Quoting vs self-tokenising.** `to_tsvector('simple', 'the HER-2/neu receptor')` indexes
  `'her' '-2' '/neu'` at consecutive positions; the quoted form expands to
  `'her' <-> '-2' <-> '/neu'` and matches, while splitting the token ourselves into `her`/`neu`
  would **not** (`'/neu'` ≠ `'neu'`). Likewise `to_tsvector('simple','½')` yields no lexeme at
  all, so an unguarded `'½'` operand would be silently dropped by Postgres — taking the
  minimum-matched-terms floor with it — which is why a term must hold a letter or decimal digit.

## Test evidence

- `cd apps/api && npx jest --maxWorkers=2 --testPathPatterns=rag` → **5 suites passed, 106 passed, 1 skipped**
- `cd apps/api && npx jest --maxWorkers=2` (full suite) → **59 suites passed, 1080 passed, 1 skipped, 0 failed**
- `cd apps/api && npm run build` → clean

New tests:

- `kb-fts-query.spec.ts` (25 tests) — the token filter, the Hinglish probe from the issue, the
  WhatsApp probe from the issue thread, the floor and its invariant, term caps, phrases,
  escaping, Devanagari queries, and the "must never eat" list.
- `kb-fts.spec.ts` (+5 tests, real Postgres via PGlite, real migrations) — the Hinglish probe
  matches and ranks the pregnancy chunk lexically; the same probe through the pre-#134 statement
  returns nothing; a natural-language English question matches on its content words; a one-word
  overlap is not a hit; quoted phrases are order-sensitive; `RagService` itself (not just the SQL)
  retrieves the pregnancy chunk; the health probe's own sentence goes through the same builder.
  The existing planner test now asserts the expression index is still used for the pairwise shape.

Note for reviewers: `apps/api/jest.config.js` ignores `/\.claude/`, so running jest from inside an
agent worktree needs `--testPathIgnorePatterns=/node_modules/ --modulePathIgnorePatterns='<rootDir>/dist/'`.
The numbers above were produced that way; on a normal checkout the plain commands apply.

## Open questions / follow-ups

- **Ranking cost.** The statement computes `to_tsvector(c.content)` again for `ORDER BY
  ts_rank_cd` over *every* matching row before `LIMIT`. That was free while the AND query matched
  nothing; now it is real work. The floor bounds it (every conjunct carries a selective term), and
  the health probe's sentence was chosen to be selective for the same reason, but the p95 of
  `hybrid_retrieval` on production should be watched after deploy. If it bites, the fix is a
  bounded candidate CTE, not a change to the builder.
- **Re-measure `ftsCount > 0`** in the `hybrid_retrieval` logs before/after, as the issue asks,
  and re-run the WhatsApp probe from the issue comment
  (`python wa_qa.py --count 1 --burst 0 --seed 1789259448`, or the same question on web).
- **Backlog P2-8** (hybrid weights are not renormalised when one arm returns nothing) is
  unchanged by this PR and still worth doing: it is what turned a silent lexical arm into pushed-
  under-the-gate evidence in the first place.
- The generic-term list (`cancer`, `tumour`, `disease`, `patient`, `treatment`, `therapy` + Hindi)
  is hand-written. A corpus-derived document-frequency cutoff would be more principled; left out
  deliberately to keep this PR code-only and reviewable.

Fixes #134


gautamgauri (commented):
**HOLD — privacy blocker.** The retrieval/FTS design is promising, but the current test fixture explicitly copies a real user's medical query into the repo. AGENTS.md forbids raw patient chat text in fixtures/tests/docs. Replace every verbatim user-transcript instance with a synthetic query that preserves the same Hinglish terms/retrieval shape, then rerun the same lexical and RagService regressions. No objection to the tsquery approach itself once that is fixed.

gautamgauri:
Codex P1 (raw user chat in a test fixture) is addressed in 32849b4.

`PROBE_A` in both `kb-fts-query.spec.ts` and `kb-fts.spec.ts` is now one **synthetic** Hinglish probe of the same shape (kinship word + "cancer hai" + "dawai se" + "bachche ko nuksaan hoga" + filler tail), and the "as the user typed it" comments are replaced with an explicit note that the string was written for the test (AGENTS.md §1.5). The same fragment was also de-quoted from the kinship-stopword doc comment in `kb-fts-query.ts`. `git diff origin/main` no longer contains any verbatim user text.

Regression intent is unchanged: the probe still translates to medicine/baby/harm/pregnant, still lexically matches and ranks `chunk-pregnancy` first through the shipped #134 statement, and still returns zero rows through the pre-#134 `websearch_to_tsquery` statement.

Expectation deltas the new wording forced (called out so they are not mistaken for behaviour changes): `affected` → `harm` (nuksaan translates to harm), `bachcha`/`affected` → `bachche`/`nuksaan` in the untranslated-terms assertion, noise list `mausi`/`ki`/`exact` → `bhabhi`/`uski`/`jaldi`, and `bhabhi`/`bhabi`/`bhabhiji` added to the Hinglish kinship stopwords (the only gap in that list the new probe hit).

Verified: `npx jest --maxWorkers=2 --testPathPatterns=rag` → 5 suites, 106 passed / 1 skipped; `npm run build` clean.

Out of scope but worth a follow-up: the same original sentence still sits on `main` in `cross-lingual.service.spec.ts:126` and as a fragment in a `cross-lingual.service.ts` comment.

gautamgauri:
Re-review on head `32849b4`: **READY**. The privacy blocker is closed on this branch: the retrieval regression now uses synthetic Hinglish text and no verbatim caregiver message remains in this PR's changed test/code surface. The retrieval behavior under test is preserved and the RAG suite/build evidence is good. Separate repo-wide cleanup of the older copy on `main` belongs to #150; I would merge that privacy containment first, but it is not a defect in this head.

Issue #91, opened 2026-09-06
Ingest OncoTalk video transcripts into the KB — feasibility and plan


gautamgauri:
## Summary

Investigated what it would take to put the SCCF YouTube videos into the knowledge base so the bot can answer from them.

**Recommendation: do not ingest the channel as a whole. Ingest at most a hand-reviewed subset, and only after #86 is cleaned up.**

The blocker is not cost (it is about 6 cents) and not volume (about 2.6% of the index). The blocker is that **two-thirds of the channel is oncologist-to-oncologist drug talk, all the captions are machine-generated and demonstrably wrong on medical words, and the folder these transcripts would land in is already configured as a top-tier trusted source that makes the bot's safety checks *less* strict.**

Three important corrections to the premise I was given, established by looking at the code and the channel:

1. **This is not third-party content.** It is SCCF's own channel (`UCI242a2_VRTCdCbpXzyeW4w`, @suchitracancercarefoundati1877). The licensing question is much smaller than expected — but it is replaced by a harder consent/attribution question, see "Open decisions".
2. **A pipeline already exists** — three of them, actually — and the main one is silently broken.
3. **The content is mostly not patient education.** It is conference recordings and pharma-sponsored clinical panels.

---

## 1. What already exists today

There are **three** separate, overlapping YouTube ingestion implementations in the repo. Nobody appears to have run any of them successfully: the target folder `kb/en/01_suchi_oncotalks/` contains exactly one hand-written file (`cancer-basics.md`) and zero transcripts, and the manifest has 1 document in that category.

| # | Where | What it is | State |
|---|---|---|---|
| 1 | `scripts/youtube-transcripts/` | Python + `youtube-transcript-api`, meant to be run monthly via `update_monthly.sh` | Has a bug (below); never run |
| 2 | `scripts/convert-youtube-to-kb.py` + `scripts/YOUTUBE_INGESTION_GUIDE.md` | Python, converts transcript JSON exported from a Google Colab notebook | Depends on a Colab notebook that is not in the repo |
| 3 | `apps/api/src/modules/youtube/` (1,114 lines) | A live admin API on production: `POST /admin/youtube/ingest` | Deployed, reachable, and broken |

### Bugs found

**(a) The live API endpoint silently produces empty documents.**
[...]

PR #144, opened 2026-09-13
feat(kb): OncoTalks transcript pipeline — guarded ingest, staged manifest, review gate (#91)

gautamgauri:
Part of #91.

Turns the three half-finished YouTube ingestion paths into one pipeline that produces **reviewable drafts** and cannot reach the live index on its own. No Prisma migration, no trust-tier change, no safety-module change, no database writes.

The transcripts themselves are in a separate, clearly-labelled draft PR (`content/91-oncotalks-staged-transcripts`, branched from this one) so nobody is asked to approve pipeline code and unverified medical content in the same click.

## What it does

- **`transcript-sections.ts` (new)** — groups timed caption segments into ~900-character sections broken at sentence ends or pauses, each under a `##` heading. `ingest-kb.ts:130` chunks on markdown headings and otherwise cuts blind every 1,200 characters, so a raw caption dump would have produced chunks that start and end mid-sentence — materially worse evidence than the NCI pages the system was built around.
- **Timestamps in the chunk text** — every section heading carries its `MM:SS` and the body carries a `?t=` deep link, so a retrieved chunk contains the means to locate the claim in the recording. This is a workaround for the missing `KbChunk.startTimeSeconds` column, not a substitute for it (§5, D5).
- **`curation.json` (new)** — all 54 videos on the channel with an explicit `include`/`exclude` decision and a written reason. 11 include, 43 exclude. Inclusion is a committed, diffable artefact, not a flag on a command line.
- **`youtube-transcript-draft.ts` (new)** — offline local tool that regenerates every draft from `curation.json`. `--check` is a read-only drift check.
- **json3 instead of VTT** — YouTube's auto-caption VTT is a *rolling* format where each cue repeats the tail of the previous one; a naive parse duplicated roughly every line and would have embedded it twice. The old code also read `<base>.vtt`, a filename yt-dlp never writes, so that whole fallback could only ever throw ENOENT — dead code that looked like a working fallback.
- **Machine-*translated* tracks are refused.** YouTube offers auto-translated captions for any video with ASR, so asking for `en` on a Hindi video silently returned a translation of a transcription — two lossy machine steps stacked on medical content. The pipeline resolves the `-orig` track explicitly.

## The empty-doc fix (§1a / §1b)

`POST /v1/admin/youtube/ingest` is deployed and reachable on production, and it wrote **empty documents into `kb/manifest.json` with `status: "active"`**.

`youtube-transcript@1.2.1` does not throw when a video has no captions — it resolves with an empty array. `[]` is truthy, so every `if (!transcriptData)` guard in the fallback chain concluded it had succeeded; yt-dlp and Puppeteer were never reached. `kb.service.ts` then wrote a markdown file with an empty body plus a manifest entry pointing at it. Nothing checked transcript length anywhere in either file.

Every such check now goes through `hasSegments()` (an explicit array-length test) plus a `MIN_TRANSCRIPT_CHARS = 400` assertion, applied at each hand-off: after the npm package, after yt-dlp, after Puppeteer, in `saveToKb`, and in `batchSaveToKb`. A rejected transcript writes **no file and no manifest entry** and is reported in `skipped` with its reason.

The second half: `kb.service.ts` resolved its output to `process.cwd()/../../kb`, which on Cloud Run is container scratch discarded on restart — so everything it wrote vanished, never reached version control, and was never reviewed, while the endpoint reported success. `resolveKbRoot()` now refuses any directory without a `manifest.json` in it, so a misconfigured deployment fails loudly.

**Containment, beyond the guards:**

| Layer | Effect |
|---|---|
| `YOUTUBE_INGEST_ENABLED` (default `false`, `false` in both Cloud Build pipelines) | The route is off. Admin credentials alone must not start writing KB documents from uncorrected machine captions. |
| Manifest writes default to `kb/manifest.oncotalks-pending.json` | `ingest-kb.ts:405` reads `<kbRoot>/manifest.json` and nothing else, so `npm run kb:ingest` **cannot** pick staged entries up. Writing the live manifest requires naming it. |
| `status: "inactive"`, `reviewStatus: "pending"` on every doc | `rag.service.ts` filters on `d.status = 'active'`; a draft cannot reach an answer. |
| `update_monthly.sh` no longer runs `npm run kb:ingest` | That path would have chunked and embedded uncorrected machine captions of medical conversation with no human in the loop. |

Also fixed: the Python extractor's `languages: [en, hi]` config was ignored (§1c) — `process_video` called `extract_transcript(video_id)` without it, so the default `["en"]` always won on a channel that is mostly Hindi. It also filed every draft under `kb/en/` regardless of the language it claimed in its own frontmatter, and defaulted `--manifest` to the live `kb/manifest.json`.

## Tests

`cd apps/api && npx jest` → **62 suites, 1111 passed, 1 skipped** (baseline before this change: 61 suites, 1102 passed). `npm run build` passes. `scripts/check_deploy_config_parity.py` passes.

**63 tests across 4 youtube suites**, including regression tests that fail against the old code:

- the empty-array return driven end-to-end through the controller, asserting **both** manifests are byte-identical afterwards and no language folder was created
- a sub-minimum transcript, skipped with a reason rather than written
- a machine-translated caption track, refused
- container scratch as `KB_ROOT`, refused (§1b)
- the flag gate: unset, `"false"`, `"1"`, `"yes"`, `""` all refuse, and refuse *before* any network call
- `updateManifest` defaults to the staging manifest and leaves `kb/manifest.json` untouched

One test-infrastructure fix rides along: `jest.config.js` matched `/.claude/` anywhere in a path, which also matched an agent worktree's *own* path — so `npx jest` inside one reported "No tests found" and looked green. It is now anchored at the checkout's own repo root; behaviour from the main tree and in CI is unchanged.

## Deferred — deliberately not in this PR

- **`KbChunk.startTimeSeconds` + `?t=` citations.** Needs a Prisma migration (§5, D5).
- **The `01_suchi_oncotalks` trust tier.** It is `priority: 'high'` today — the same tier as NCI — which multiplies relevance 1.5x **and** sets `hasTier1 = true` in `evidence-gate.service.ts:215`, dropping the evidence requirement to a single passage from a single source and switching off the abstention check. `trusted-sources.config.ts` is untouched here on purpose: changing it alters safety behaviour and needs SCCF sign-off (D2).
- **Decisions D1–D5** — whether machine-transcribed clinician talk should be retrievable by a patient bot at all (D1); trust tier (D2); speaker consent and pharma-sponsorship attribution (D3); whether machine captions are acceptable uncorrected (D4); timestamped citations as a hard prerequisite (D5).
- **Ingestion itself**, which §6 sequences after the #86 duplicate cleanup so there is a trustworthy baseline to measure against.


gautamgauri (commented):
**HOLD — pipeline containment is good, but the review workflow is not yet reliable.** Three issues on the current head need fixing: (1) the included English draft is generated under ignored `kb/en/**`, so the tracked pending manifest can reference a file that ordinary git review never sees; (2) `--only` rewrites the whole staging manifest with the subset and can delete the other entries; (3) `--check` treats a missing expected draft as clean drift because `before === null` is not recorded as a mismatch. Put English drafts in a tracked review path (or unignore them), merge `--only` results into the existing manifest, and make missing drafts fail `--check`. The inactive/pending/live-manifest separation is otherwise the right architecture.

gautamgauri:
All three Codex P2s in `apps/api/src/scripts/youtube-transcript-draft.ts` are addressed.

**English drafts under an ignored path** — `a51b80d`. `.gitignore` now excludes `kb/en/*` (the directory's *contents*) instead of `kb/en/`, because git does not descend into an excluded directory and would never reach a `!` line under it, then un-ignores `!kb/en/01_suchi_oncotalks/`. Verified with `git check-ignore -v`: a path under the OncoTalks directory is not ignored, `kb/en/02_nci_core/…` still is. `kb/README.md` documents the exception and why it is shaped that way. yt-dlp and network were available, so the English draft was actually generated (`en-orig`, 16 sections, 19,193 chars) and committed on the stacked content branch — `#145` now carries all 11 entries with `"tracked": true` and no entry pointing at a file git cannot see.

**`--only` wiping the manifest** — `9302e64`. `selectVideos()` resolves `--only` *before* anything is written and throws on an id that is not in `curation.json`, or one curated as `exclude`; the CLI wrapper turns that into a non-zero exit with no manifest write. `mergeStagedManifest()` folds results in by id — a `--only` run replaces just that entry (appending when new), a full run still rewrites the list — and merges any existing entry under the generated one, so hand-maintained fields like `tracked` survive regeneration. Confirmed live: `--only 6edsfQY1TZU` against the real staging manifest left all ten Hindi entries intact (11 total), and `--only NOSUCHID123` exits 1 with the manifest byte-identical.

**`--check` passing on a missing draft** — `34df537`. `classifyDraft()` returns `"missing" | "differs" | "match"`; `before === null` is drift, and the failure output labels each path so an edit is distinguishable from an absence. `--check` stays read-only either way. Confirmed live: moving the committed draft aside makes `--check` exit 1 ("not committed"), restoring it exits 0 with a clean tree.

`main()` is exported and self-invokes only under `require.main === module`, so all of this is tested end to end against a temp KB checkout with the transcript fetch stubbed. 23 tests in the new `youtube-transcript-draft.spec.ts` (14 for the `--only` merge and id validation, 9 for the drift classifier and `--check`); 86 across `--testPathPatterns=youtube`, all green, `npm run build` clean.

`#145` was rebased onto `34df537` with `--force-with-lease` (it applied cleanly — only new commits below it) and its title/body updated: it is 11 transcripts now, not 10, and the paragraph saying the English draft "cannot be committed" is gone.

gautamgauri:
Re-review on head `34df537`: **READY**. The three review-workflow blockers are closed: the English OncoTalks draft path is explicitly tracked while the rest of `kb/en` remains ignored; `--only` validates before writes and merges a single result without truncating the staging manifest; and `--check` now treats a missing expected draft as drift while remaining read-only. The inactive/pending/live-manifest separation remains intact. #145 should stay draft for SCCF content review.

PR #146, opened 2026-09-13
fix(db): unique (docId, chunkIndex) on KbChunk + ingest preflight (#86)

gautamgauri:
## The cleanup is already applied — this is only the guard

The duplicate DELETE + `VACUUM` ran on production on **2026-09-06**. Re-verified read-only on **2026-09-13** with query (A) of `scripts/sql/kb_duplicate_cleanup.sql`:

| metric | 2026-09-06 findings | today |
|---|---|---|
| KbChunk rows | 73,802 | **48,737** |
| non-deterministic-id rows | 25,065 | **0** |
| duplicate `(docId, chunkIndex)` rows | 25,065 | **0** |
| docs | 1,439 | **1,439** |

**This PR touches no data.** It ships the prevention layer from §7 of the findings, which was explicitly deferred until after the cleanup because the migration would have failed against the old rows.

## What it prevents

`ingest-kb.ts` upserts on the deterministic id `docId::chunk::N`. A run that wrote Prisma `@default(uuid())` ids left rows the upsert can never reach, so every later run inserted a **byte-identical second copy** of each chunk instead of overwriting it: 25,065 extra rows, 34% of the index, across 511 of 1,439 docs.

Nothing errored. The bill was paid silently in retrieval: the duplicate embeds identically, scores identically, and lands next to its twin in the neighbour set, so **91% of eval queries built their answer on ~4.1 distinct chunks instead of 6**, and **71% lost an entire source document** from their evidence set. After this PR a recurrence is a `23505 unique_violation` at ingest time.

## Changes

- **`schema.prisma`** — `@@unique([docId, chunkIndex])` on `KbChunk`.
  This makes the existing non-unique **`KbChunk_docId_idx` largely redundant** (the new key's leading column is `docId`, so a `docId`-only lookup can use it). It is **deliberately left in place**: dropping an index that retrieval and the #83 prune path both touch belongs in its own change with its own before/after measurement, not bundled into a safety guard.
- **`prisma/migrations/20260913000000_kb_chunk_unique_position/`** — one `DO` block (Prisma 5 does not wrap a migration file in an implicit transaction, so a single block is what guarantees a raise leaves nothing behind). It **RAISEs with the actual row counts and a pointer to `scripts/sql/kb_duplicate_cleanup.sql`** if duplicate positions exist, instead of failing with a bare `23505` naming two row ids — which in a Cloud Build migration job reads as an unexplained deploy failure. Idempotent: skips when the index is already present and valid.
  Locking note in the header: a btree over `(text, int)` on ~49k rows takes a `SHARE` lock for seconds and blocks only writes (the hand-run ingest job), unlike the GIN-over-`content` build in `20260908000000` that had to go out `CONCURRENTLY`.
- **`ingest-kb.ts`** — preflight that refuses to run when any `KbChunk` id is not `docId::chunk::N`, pointing at #86 and the cleanup script. This catches the precursor the unique index *cannot* see: a legacy-id row sitting **alone** at its position is still invisible to the upsert and is exactly what becomes a duplicate the moment the chunker's output shifts. A `--dryRun` warns instead of exiting. The verdict is a pure function in `kb-index-preflight.ts` over the same counting query as `OpsMetricsService.kbIndexIntegrity` / `npm run ops:metrics`.
- **Migration pins** — `MIG` in `cloudbuild.gated.yaml` and the fallback default in `apps/api/scripts/migrate-with-repair.sh`, plus `docs/cloudbuild-gated-issues.md`. `scripts/check_deploy_config_parity.py` passes (21 env vars, 19 secrets, both pipelines identical). No other pin exists in `scripts/`.
- **`migrate-with-repair.sh` step (6)** — asserts `KbChunk_docId_chunkIndex_key` exists and is valid. It runs **unconditionally**, not only on the repair branch, so a migration that silently did not apply — or that the repair path wrongly marked applied, which for an index-creating migration would record something untrue — fails the job instead of deploying green. This is partial progress on Issue 3 of `docs/cloudbuild-gated-issues.md`.

## Manual apply steps (production)

`prisma migrate dev` fails against prod (P3014 — `suchi_app` has no shadow-DB permission), so use the documented diff + psql + resolve pattern.

```bash
# 0. Backup first
gcloud sql backups create --instance=suchi-db \
  --project=gen-lang-client-0202543132 --description="pre-86-unique-position"

# 1. Cloud SQL proxy (terminal 1)
cloud-sql-proxy gen-lang-client-0202543132:us-central1:suchi-db \
  --credentials-file ~/.config/gcloud/legacy_credentials/suchi-scheduler@gen-lang-client-0202543132.iam.gserviceaccount.com/adc.json \
  --port 5432

# 2. Terminal 2 — DATABASE_URL from Secret Manager
export DATABASE_URL="$(gcloud secrets versions access latest --secret=database-url)"

# 3. Confirm the table is still clean — expect  48737 | 0 | 0
psql "$DATABASE_URL" -c "SELECT count(*)::int AS total_rows,
  (count(*) FILTER (WHERE id NOT LIKE '%::chunk::%'))::int AS non_deterministic_id_rows,
  (count(*) - count(DISTINCT (\"docId\", \"chunkIndex\")))::int AS duplicate_position_rows
  FROM \"KbChunk\";"

# 4. Apply. If duplicates are back it RAISEs and commits nothing.
psql "$DATABASE_URL" \
  -f apps/api/prisma/migrations/20260913000000_kb_chunk_unique_position/migration.sql

# 5. Verify the index is there, unique and valid
psql "$DATABASE_URL" -c "SELECT indexrelid::regclass, indisunique, indisvalid
  FROM pg_index WHERE indexrelid = 'public.\"KbChunk_docId_chunkIndex_key\"'::regclass;"

# 6. Record it as applied so migrate deploy skips it
cd apps/api && npx prisma migrate resolve --applied 20260913000000_kb_chunk_unique_position
npx prisma migrate status
```

The migration is idempotent, so step 4 is safe to re-run, and a later `prisma migrate deploy` from the gated pipeline is a no-op.

Optional, unrelated to this PR: the HNSW index is still 182 MB — `VACUUM` does not shrink an HNSW graph, so reclaiming the ~34% needs `REINDEX INDEX CONCURRENTLY kbchunk_embedding_hnsw_idx` in a quiet window (memory-sensitive on the f1-micro). Not run, still needs a decision.

## Tests

`kb-index-preflight.spec.ts` — 6 cases over the verdict function (clean prod shape, empty table, the January 2026 damage, a lone legacy-id row, duplicates with clean ids, bigint coercion), plus the migration **replayed against a real Postgres** (PGlite, in-process WASM; never the production database): it creates the index on clean rows, the table then rejects a duplicate with `23505`, it refuses with an actionable message on dirty data leaving no half-built index behind, and it is idempotent. A migration no test has ever executed is only a claim — #92 is the counter-example.

```
apps/api: 59 suites, 1060 passed, 1 skipped, 0 failed
prisma validate            OK
prisma generate            OK
npm run build              OK
check_deploy_config_parity OK
```

Part of #86


gautamgauri (commented):
**HOLD — two guardrail bugs before merge/migration.** First, the ingest preflight only checks whether an ID contains `::chunk::`; it must compare the exact deterministic value `${docId}::chunk::${chunkIndex}` or corrupt legacy IDs can pass preflight and fail later as an opaque unique violation. Second, the new migration post-check is inside `set -e`, so a transient DB error exits before the advertised retry loop can run. Fix the exact-ID predicate and capture/retry the index-check status safely. After that, the unique constraint + preflight design is sound; only then should the production migration be applied.

gautamgauri:
Both P2s addressed.

**1. `migrate-with-repair.sh` — errexit killed the retry loop** (`da328d2`)

Step (6) ran `node` under `set -e`, so the first non-zero exit terminated the job before `kb_idx_status=$?` and before any of the three advertised retries. Wrapped the loop in `set +e` / `set -e`, matching step (2)'s status-check loop, and named the attempt/delay values. Exit 0 (valid) and exit 2 (genuinely missing) stay verdicts and break immediately; only exit 1 (driver/connection error) retries — 3 attempts, 2s apart.

No bats/shell harness in the repo, so verified with `bash -n` + `sh -n` and a stubbed `node` on PATH:

```
===== AFTER FIX (transient failure on attempt 1) =====
=== (6) Verify KbChunk unique position index (issue #86) ===
[stub node] KbChunk check attempt 1 -> simulated Cloud SQL connection error
Attempt 1 failed (exit 1), retrying in 2s...
[stub node] KbChunk check attempt 2 -> KbChunk unique position index present and valid.
Migration job completed successfully.
EXIT=0

===== BEFORE FIX (same stub, pre-fix script from git) =====
=== (6) Verify KbChunk unique position index (issue #86) ===
[stub node] KbChunk check attempt 1 -> simulated Cloud SQL connection error
EXIT=1
```

The two failure paths still fail, correctly:

```
===== index genuinely missing (exit 2): fails fast, no retries =====
[stub node] attempt 1 -> exit 2
❌ KbChunk unique position index check failed!    EXIT=1

===== persistent connectivity error (exit 1 x3) =====
[stub node] attempt 1 -> exit 1 ... attempt 3 -> exit 1
All 3 attempts failed (last exit: 1)
❌ KbChunk unique position index check failed!    EXIT=1
```

**2. `kb-index-preflight.ts` — substring test scored unreachable rows as deterministic** (`3f19c66`)

Replaced `id NOT LIKE '%::chunk::%'` with an exact comparison, extracted as `NON_DETERMINISTIC_ID_PREDICATE`:

```sql
id IS DISTINCT FROM ("docId" || '::chunk::' || "chunkIndex")
```

`IS DISTINCT FROM` rather than `<>` so a NULL in any of the three columns counts as an offender instead of dropping out of the `FILTER`. The refusal now names up to 5 offending ids plus a `+N more` tail; the samples come from their own `LIMIT 5` subquery rather than a sliced `array_agg`, so a table in the January-2026 state doesn't materialise 25,065 strings to print five.

Consistency of the same predicate elsewhere — checked all five sites:

| Site | Action |
|---|---|
| `OpsMetricsService.kbIndexIntegrity` | aligned to exact |
| `src/scripts/ops-metrics.ts` | aligned to exact |
| migration's diagnostic counter | aligned to exact |
| `kb_duplicate_cleanup.sql` query (A) + post-check | aligned to exact |
| `kb_duplicate_cleanup.sql` sections 2–4 | **deliberately left on the substring form**, now documented |

Sections 2–4 aren't a health question — they select the DELETE cohort (legacy uuid-id rows inside docs that also hold deterministic rows), the exact set whose "identical surviving twin" invariant was verified offline at 25065/25065 on 2026-09-06. Tightening it would silently widen a destructive DELETE past what was verified. The documented consequence: an off-shape id in a doc with *no* deterministic rows keeps the health signal non-zero after the cleanup. That's correct — it needs its own decision (re-ingest the doc, or delete the row), not a blanket DELETE.

**Tests** — 7 new cases: wrong-doc id (`doc-other::chunk::3` on doc-screening/3), non-numeric suffix (`legacy::chunk::x`), sample capping, the `+N more` tail and its absence, and a row missing the sample column entirely. Plus two PGlite cases running the shipped SQL against the real schema, one asserting it catches all three ids the substring form called healthy (including a suffix that drifted from its own `chunkIndex`). Both reuse the shared database — no new PGlite instance.

```
Test Suites: 3 passed, 3 total
Tests:       44 passed, 44 total   (was 37)
```

`npm run build` clean.

gautamgauri:
Re-review on head `3f19c66`: **READY**. Both guardrail findings are closed. The migration post-check now actually survives a transient exit under `set -e` and retries only connectivity/driver failures; the ingest integrity predicate now compares the exact deterministic id (`docId::chunk::chunkIndex`) with NULL-safe semantics and is aligned across the related diagnostics. The unique-position design is ready to merge. **Operational sequencing remains important:** apply the production migration through the documented proxy/manual flow only after this PR is merged; do not let a normal deploy be the first prod application.

Issue #27, opened 2026-06-04
Social: Configure LinkedIn org page posting


gautamgauri:
## Problem
LinkedIn posting is currently disabled. The existing token (`LINKEDIN_ACCESS_TOKEN`) only has personal posting scope (`w_member_social`) — it cannot post to an org page.

## What's needed
1. Create/confirm the Suchi Cancer Care Foundation LinkedIn org page
2. Re-authenticate LinkedIn OAuth with `w_organization_social` scope to get a new token
3. Get the org page ID and set `LINKEDIN_AUTHOR_URN=urn:li:organization:ORG_ID`
4. Add both `LINKEDIN_ACCESS_TOKEN` and `LINKEDIN_AUTHOR_URN` back to `cloudbuild.yaml` secrets
5. Create the secrets in GCP Secret Manager and deploy

## Current state
- `LINKEDIN_ACCESS_TOKEN` and `LINKEDIN_AUTHOR_URN` are intentionally removed from `cloudbuild.yaml` (posting to personal account was disabled)
- Token expires ~July 20 2026 — new token will be needed regardless
- Code is in `apps/api/src/modules/admin/social-post.service.ts` — `postLinkedIn()` method is ready, just needs credentials

gautamgauri:
## Triage: **KEEP OPEN, but rescope it to a yes/no decision**

Nothing in the checklist has been done, and the situation has got simpler rather than more complicated.

### Current state, verified today

- `LINKEDIN_ACCESS_TOKEN` and `LINKEDIN_AUTHOR_URN` appear in **neither** `cloudbuild.yaml` nor `cloudbuild.gated.yaml`, so they are not on the live service. Steps 4 and 5 of this issue are untouched.
- The code path is intact but inert: `postLinkedIn()` (`apps/api/src/modules/admin/social-post.service.ts:359-362`) returns early with `not_configured`, and `liConfigured` (`:412`) is `false`, so LinkedIn is simply omitted from the approval email (`:460`). **Nothing is broken** — the feature is absent, not failing.
- The 60-day token expired around **2026-07-20**, seven weeks ago. So step 2 ("re-authenticate with `w_organization_social`") is now mandatory rather than optional: there is no token left to salvage, and this would be true even if the org page appeared tomorrow.

### Why it hasn't moved

[...]

PR #147, opened 2026-09-13
feat(social): LinkedIn organisation-page posting via the versioned Posts API

gautamgauri:
## What this does

Founder decision on #27: **SCCF has a LinkedIn organisation page**, so path (a) — re-enable org-page posting rather than delete the code.

This is the engineering half. After merge, the only thing standing between Suchi and a live LinkedIn post is a human browser sign-in (the list is at the bottom).

### 1. `postLinkedIn()` now actually targets an organisation page

`apps/api/src/modules/admin/social-post.service.ts`

- **Endpoint migrated** from the deprecated `/v2/ugcPosts` UGC API to the versioned Posts API, `POST https://api.linkedin.com/rest/posts`, with both required headers (`LinkedIn-Version: 202608`, `X-Restli-Protocol-Version: 2.0.0`). The old `specificContent`/`com.linkedin.ugc.ShareContent` body shape is gone; the new body uses `commentary` + `content.article` + `distribution.feedDistribution: MAIN_FEED`. The Posts API does not scrape URLs, so the article title is now passed through from the draft and sent explicitly.
- **Author** is the `urn:li:organization:<id>` from `LINKEDIN_AUTHOR_URN`, posting as the page with a `w_organization_social` token. A malformed URN (e.g. a bare numeric id) is rejected *before* the network call, with a log line saying what it saw.
- **Commentary escaping.** LinkedIn "little text" reserves `\ | { } @ [ ] ( ) < > # * _ ~` and requires them escaped "even if those characters are not used in one of the supported elements" — an unescaped `(` in Gemini-written copy is a 422. New exported `escapeLittleText()` handles it. Side effect worth knowing: hashtags in LinkedIn copy now render as literal text, not links.
- **Failures are legible.** A 401 logs `access token expired or revoked … see docs/LINKEDIN_ORG_POSTING.md` and returns `token_expired (HTTP 401) — see docs/LINKEDIN_ORG_POSTING.md`; a 403 names the missing `w_organization_social` scope or page-admin role; a network error is caught instead of escaping as an unhandled rejection. Previously every failure was a bare `HTTP 401`, which is part of why this integration sat dead for months.

`LINKEDIN_ACCESS_TOKEN` and `LINKEDIN_AUTHOR_URN` were already `z.string().optional()` in `env.validation.ts` — confirmed, unchanged. With them absent the method still returns `not_configured` and the approval email omits LinkedIn (`liConfigured`), so merging this changes nothing about current behaviour.

### 2. Deploy config

Both secrets added as `--set-secrets` entries to **`cloudbuild.yaml` and `cloudbuild.gated.yaml`** — `--set-secrets` replaces rather than merges, and adding to one file only is how the Meta credentials were dropped in June. `python3 scripts/check_deploy_config_parity.py` → `OK: all deploy pipelines carry identical Cloud Run runtime config` (21 env vars, 21 secrets on both).

⚠️ **This makes the two Secret Manager secrets a deploy-time dependency.** Cloud Run resolves every secret reference when the revision is created, so `linkedin-access-token` and `linkedin-author-urn` must exist before the next `suchi-api` deploy — a placeholder value is fine, a missing secret fails the deploy. No GCP secrets were created and nothing was deployed from this branch.

### 3. Runbook — `docs/LINKEDIN_ORG_POSTING.md`

Prerequisites (Community Management API product, page-admin member, registered redirect URL), finding the org id and building the URN, the exact OAuth URL with `w_organization_social` (+ `r_organization_social`, optional, only for the org lookup), the code exchange, `gcloud secrets create` / `versions add` under the names `cloudbuild.yaml` references, deploy, verification via the approval email, the 401/403/422 log signatures, the 60-day expiry, and how to turn LinkedIn back off. `docs/RELIABILITY_BACKLOG.md` P2-6 records the decision and points at it.

**Refresh tokens:** documented, with the caveat that decides it — programmatic refresh is only issued to approved Marketing Developer Platform partners. A standard Community Management app gets no `refresh_token` and must be re-authorised by hand every 60 days. The helper script reports which case this app is in, and the doc has the refresh curl for the event that it is refresh-eligible.

### 4. `scripts/linkedin-oauth-exchange.ts` (new — nothing existing to reuse)

No OAuth-exchange script existed under `scripts/` or `apps/api/src/scripts/`. This one is dependency-free (node builtins + global fetch) with three subcommands: `url` (print the authorisation URL), `exchange --code` (swap code for token; reports granted scopes, exact expiry date, and whether a refresh token came back), `orgs` (list the organisation URNs this token may post for, so the author URN is copied rather than hand-assembled). Wired as `npm run linkedin:auth`. Tokens are printed for piping into `gcloud secrets versions add` and never written into the repo.

## Testing

- `npx jest --testPathPatterns=social` → **18 passed, 1 suite**, including 8 new tests: not-configured short-circuit (no network call), the versioned-endpoint contract (URL, all three headers, org author, article content, no `ugcPosts`, no `specificContent`), the 401 runbook-pointer message, the 403 scope message, malformed-URN rejection before the call, network-error containment, and two `escapeLittleText` cases.
- `npm run build` (apps/api) → clean.
- `python3 scripts/check_deploy_config_parity.py` → OK.
- No KB, prompt, safety, or clinical-policy files touched. The social safety gate and the `safetyBlocked` hard block are untouched and still covered by the pre-existing tests in the same suite.

## Human steps that remain (all require a browser or `gcloud`)

1. Confirm the SCCF LinkedIn **company page**, and that a LinkedIn member account you control is an **ADMINISTRATOR** of it (prefer a shared ops account over a personal one — the token dies with the member's page role).
2. In the LinkedIn developer app: set its Company to the SCCF page, request and get the **Community Management API** product approved (`w_organization_social` does not appear on the consent screen until then), and register a redirect URL on the Auth tab.
3. Run `npm run linkedin:auth -- url`, open the URL as the page admin, approve, copy the `code`, then `npm run linkedin:auth -- exchange --code <code>` — note the printed expiry date and whether a refresh token came back.
4. Get the org id (`npm run linkedin:auth -- orgs`, or read it off `linkedin.com/company/<id>/admin/`) and build `urn:li:organization:<id>`.
5. **Before the next deploy:** `gcloud secrets create linkedin-access-token` and `linkedin-author-urn`, then add a version to each (exact commands in the runbook; use `printf %s`, not `echo` — a trailing newline in the token is a confusing 401).
6. Deploy `suchi-api` via Cloud Build from a clean tree and promote the new revision to 100% traffic.
7. Verify: the next social approval email should show a LinkedIn block, a "LinkedIn only" button and "Approve all (3 platforms)"; click LinkedIn only on a real draft and check the page feed.
8. Create a recurring **55-day calendar reminder** to rotate the token (steps 3 → 5 → 6, ~10 minutes). Nothing in the system can do this for you unless LinkedIn grants MDP partner status.

Do not merge before step 5 is possible — merging is safe, but the *next deploy after merge* needs both secrets to exist.

Part of #27.


gautamgauri (commented):
**HOLD — security blocker.** The OAuth exchange helper currently prints live access/refresh tokens to stdout. That violates the repo rule against secrets in logs and makes terminal recording/redirection an easy credential leak. Change the flow so tokens are passed directly to Secret Manager or otherwise handled without echoing raw values. Also make the org lookup include every page role the runbook says is supported (e.g. CONTENT_ADMIN), not only ADMINISTRATOR. Do not create the new Cloud Run secret references or deploy this branch until the token-handling path is safe.

gautamgauri:
Both Codex findings addressed.

**P1 — tokens printed verbatim to stdout** (`0cd215a`)
`exchange` never prints a token now. It requires exactly one destination and exits 1 with none, instead of falling back to printing:
- `--to-secret-manager <name>` — pipes the value into `gcloud secrets versions add <name> --data-file=-` over the child process's stdin (no argv, so nothing in `ps`; no shell interpolation; no temp file) and prints only the new version name. `--secret-project <id>` overrides the project.
- `--out <path>` — writes it with mode 0600 and no trailing newline; refuses to clobber an existing file unless `--force` (which re-applies 0600).
- `--print-token` — the escape hatch, explicit opt-in, prints a warning first.

A returned refresh token goes to a sibling destination (`<secret>-refresh` / `<path>.refresh`). What is printed is metadata only: granted scopes, token length, days to expiry and the expiry date, and whether a refresh token came back. API/gcloud error bodies go through `redactSecrets()` first. `orgs` gained `--token-file <path>` so the token need not be exported into the environment. Argument handling is a real parser now (`--k v` and `--k=v`, unknown flags and missing values rejected).

**P2 — `orgs` filtered to ADMINISTRATOR** (`29a7b1c`)
The query goes out unfiltered, with a fallback to one query per role in `POSTING_ROLES` (ADMINISTRATOR, CONTENT_ADMIN, DIRECT_SPONSORED_CONTENT_POSTER) if an API version rejects an unfiltered `roleAssignee` query. Results collapse to one row per organisation with every role and state shown next to the URN; `organizationalTarget` is accepted alongside `organization`.

**Tests.** No jest project reaches `scripts/` (`apps/api/jest.config.js` roots at `apps/api/src`), so the pure helpers carry a `--self-test` mode — 54 checks over argument parsing, destination resolution, the summary formatter (including an assertion that it never contains the token value), redaction, the org/role table and the 0600 file write. Wired into the "Build + config parity" CI job as `npm run linkedin:auth:self-test`.

Verified locally: self-test 54/54; `npx jest --testPathPatterns=social` 18/18 in 1 suite; `npm run build` clean; `check_no_secret_literals.py` and `check_deploy_config_parity.py` both OK.

`docs/LINKEDIN_ORG_POSTING.md` steps 1/2/3/5/6 updated; the manual `printf %s | gcloud secrets versions add` path stays documented as the alternative for `--out` / `--print-token`.

gautamgauri:
Re-review on head `29a7b1c`: **HOLD — one security issue remains.** The default exchange path is now much safer and the role lookup finding is fixed, but `--print-token` still deliberately routes the live access/refresh token to stdout (`Destination { kind: "stdout" }` → `console.log(value)`). That preserves the exact terminal-recording/log-capture exposure the original P1 was about and conflicts with the repo's no-secrets-in-logs boundary. Since `--to-secret-manager` and mode-0600 `--out` already cover the legitimate workflows, please remove the stdout destination/flag entirely (including usage/self-test/docs) rather than keeping a warned escape hatch. After that, this looks ready. Separate deployment prerequisite remains: the two referenced Secret Manager secrets must exist before the first deploy after merge.

gautamgauri:
Both blockers cleared. Rebased onto `main` @ 6f57782 (linear, no merge commit).

## Blocker 1 — the rebase

Four of the five overlapping files auto-merged; one conflicted.

**`.github/workflows/deploy-api.yml` — real conflict, both sides kept.** `main` added the `check_image_assets.py` step (#140) at the same insertion point where this branch adds the Node setup + `linkedin:auth:self-test` steps. Both now run, `main`'s first. Neither side dropped.

**Auto-merged, `main`'s side preserved verbatim:**
- `social-post.service.ts` / `.spec.ts` — `main`'s `SUCHI_SITE_URL` default fix (`suchicancercare.org` → `suchitracancercare.org`, #155) is intact.
- `cloudbuild.yaml` / `cloudbuild.gated.yaml` — `main` edited the `--set-env-vars` line (domain fix + `YOUTUBE_INGEST_ENABLED=false`) while this branch edited the adjacent `--set-secrets` line. Both changes are present in both files, plus `main`'s `MIG=20260913000000_kb_chunk_unique_position` bump in the gated pipeline.

One follow-on: the branch had introduced **four fresh references to the dead domain** that `main` never saw, so they survived the merge — the helper's default `REDIRECT_URI`, two spots in the runbook, and four spec literals. The redirect one has teeth (it must byte-match the URL registered on the LinkedIn app), so all are now on `suchitracancercare.org`.

## Blocker 2 — `--print-token` is gone

Removed, not deprecated and not gated:

- dropped from `FLAG_SPEC`, so it is now an **unknown flag and an error**, not a silently-ignored one;
- the `{ kind: "stdout" }` variant is **removed from the `Destination` type**, so a printing sink is unrepresentable rather than merely unreachable — `deliver()` has exactly two branches, Secret Manager and a 0600 file;
- gone from `resolveDestination()`, its refusal text, `usage()`, the `exchange` usage line, and the runbook.

Two leaks that would have survived just deleting the flag are closed too:
- the delivery-failure handler printed `(err as Error).message` raw; a storage backend can echo what it was handed, so it now goes through `redactSecrets()` with both token values first;
- `addSecretVersion()` printed gcloud's stdout unscrubbed on the **success** path (the failure path already redacted).

## How "no token path remains" was verified

Whole file, not just the flag's branch:

1. **Structural** — `Destination` has no terminal variant; `deliver()` has two non-printing branches; `FLAG_SPEC` has no `print-token`.
2. **Every output call audited by hand** — all 42 `console.*` calls in the operational region. The ones carrying variable data are `summarizeToken` (metadata only), `redactSecrets(...)` at all three error sites, a file path, an org-ACL row, and gcloud's version name. None can carry a token.
3. **Source guard, `findTokenPrintingCalls()`** — re-derives from the file's own source on every CI run that no `console.*` / `process.std{out,err}.write` call is handed a token-bearing expression. It is proven non-vacuous: the self-test runs it against a planted `console.log(data.access_token)` and requires it to fail, and asserts the scanned region still contains the real command implementations so a moved marker can't make it pass over nothing.
4. **Runtime capture** — the self-test drives the real `deliver()` with stdout, stderr and all four `console` methods captured, asserting a fake token appears in none of: successful file delivery, failed delivery, an echoing storage backend, the exchange summary, and the usage text. Thrown exception messages are captured too.

Empirically:

```
$ ts-node scripts/linkedin-oauth-exchange.ts exchange --code FAKECODE --print-token
Unknown flag: --print-token

$ ts-node scripts/linkedin-oauth-exchange.ts exchange --code FAKECODE
No token destination given. Choose exactly one:
  --to-secret-manager <secret-name>  ...
  --out <path>                       ...
There is no option to show the token: it would be captured by scrollback and logs.
```

The supported replacement is `--to-secret-manager <name>`; the runbook now says so and carries a short note on why the flag was removed, so it doesn't get added back.

## P2 role check — holds

`POSTING_ROLES` = `ADMINISTRATOR`, `CONTENT_ADMIN`, `DIRECT_SPONSORED_CONTENT_POSTER`, used by the `organizationAcls` per-role fallback. Consistent with `printUrl()` ("ADMINISTRATOR (or CONTENT_ADMIN)"), the runbook, and the 403 message in `social-post.service.ts`. A self-test case asserts a `CONTENT_ADMIN`-only org survives `formatOrgRows`.

## Deploy prerequisites

No undeclared env var: the service reads exactly `LINKEDIN_ACCESS_TOKEN` and `LINKEDIN_AUTHOR_URN`, both `z.string().optional()` in `env.validation.ts` and both mounted as `--set-secrets` in **both** pipelines. `check_deploy_config_parity.py` → OK (22 env vars, 21 secrets on both).

⚠️ **Both Secret Manager secrets must exist before the next `suchi-api` deploy**, not before merge. Cloud Run resolves every `--set-secrets` reference when the revision is created, so a missing `linkedin-access-token` or `linkedin-author-urn` fails the deploy outright — a placeholder value is fine:

```
printf %s 'placeholder' | gcloud secrets create linkedin-access-token --data-file=- --project=gen-lang-client-0202543132
printf %s 'placeholder' | gcloud secrets create linkedin-author-urn  --data-file=- --project=gen-lang-client-0202543132
```

## Verification

- `npm run linkedin:auth:self-test` → **SELF-TEST PASSED** (69 checks, 0 failures)
- `npx jest --testPathPatterns=social` → **18 passed, 1 suite**
- `tsc --noEmit --strict` on the helper → clean
- `check_deploy_config_parity.py` / `check_no_secret_literals.py` / `check_image_assets.py` → all OK

No KB, prompt, safety or clinical-policy files touched.

PR #148, opened 2026-09-13
feat(chat,data): rank by nearest relevant centre — geocode the directory, order by capability then distance (#103)

gautamgauri:
Part of #103. **Stacked on #99** — base is `fix/hospital-geographic-fallback-north-bihar`, so review that first; this diff is only the five commits on top.

Implements the minimal sketch from the #103 data-gap assessment: geocode offline, commit the result, then rank by capability and real distance instead of by state.

## 1. Geocoding — `scripts/geocode-hospitals.py`

Offline and one-off. The API never calls a geocoder, so there is **no runtime key, no runtime latency, no runtime failure mode, and no new env var** to keep in sync across the two cloudbuild files (AGENTS.md §3). Public Nominatim, 1 req/s, descriptive User-Agent naming SCCF/Suchi with a contact URL.

Two **hard gates**, either of which writes no coordinate at all:
1. the returned address must contain the record's own `city` **and** `state` (case-insensitive, with a table of common Indian spelling variants — Bengaluru/Bangalore, Odisha/Orissa, Prayagraj/Allahabad, …);
2. the point must fall inside India (lat 6.5–37.5, lon 68–97.5).

Failures append to `scripts/geocode-unresolved.json`, so **every record either carries coordinates or is listed there — no silent third state**. Idempotent; `--force` re-geocodes.

### Counts

| | resolved | unresolved |
|---|---|---|
| hospitals | **83 / 83** | **0** |
| INDIAN_CITIES | **45 / 45** | **0** |

**Unresolved list: empty.** `geocode-unresolved.json` ships with `hospitals: []` and `cities: []`.

Hospital confidence tiers — which query answered, not how precise the fix is: `name_address` 3, `address` 33, `city` 47.

> Note on the city count: #103 says 46 entries. The table has **45**; the 46th `canonical:` in the file is the `CityEntry` interface's own field declaration.

### Gate failures encountered

No record failed the final run. Two precision bugs were found and fixed *before* it, both of which move an origin city 15–20km and so reorder the centres ranked from it:

- **District centroids beating towns.** `"Darbhanga, Bihar, India"` returns the **district** (`addresstype: state_district`) at `26.083, 86.032` ahead of the **city** at `26.157, 85.900`. Gate-passing hits on the locality tier are now ranked `city > town > … > county > state_district`.
- **Towns Nominatim will not return at all** for the plain query. `"Buxar, Bihar, India"` yields only district and tehsil boundaries; the town node appears only when the district is named too (`"Buxar, Buxar, Bihar, India"`). Most Indian district headquarters share their district's name, so that is a general second tier. City queries now pool both and take the most local result. Buxar moved `25.5045, 84.1098 → 25.5716, 83.9730`; Kishanganj `26.2984 → 26.1014`.

Sanity-checked against the six hand-checked Bihar values in #103 — all agree:

| city | expected | geocoded |
|---|---|---|
| Patna | 25.61, 85.14 | 25.6093, 85.1235 |
| Darbhanga | 26.15, 85.90 | 26.1570, 85.8995 |
| Muzaffarpur | 26.12, 85.39 | 26.1183, 85.3858 |
| Bhagalpur | 25.25, 87.00 | 25.2495, 86.9828 |
| Purnia | 25.78, 87.47 | 25.7774, 87.4731 |
| Gaya | 24.79, 85.00 | 24.7964, 85.0080 |

`apps/api/data/hospitals.json` is a **symlink** to `apps/landing/src/content/hospitals.json` — one file, one diff.

## 2. Ordering: capability first, then distance

1. **Capability, as a HARD filter, resolved before any geography.** Unlike every other filter in `searchHospitals` it does **not** degrade gracefully: a centre that cannot serve the need is dropped even if that empties the set (logged `hospital_capability_filter_empty`). Offering nothing is recoverable; routing a radiotherapy patient to a surgery-only centre is not.
2. **Distance ascending**, haversine from the patient's geocoded city.
3. **Directory `score` as tiebreak.** Several centres share a city and so share a distance to the decimal; without this the order within a city would be file order, and a tier-A TMC unit could fall below a tier-C private hospital on the same street.

Hospitals without coordinates sort **last**, never first — an unknown distance is not a short one.

**The state rungs are gone on this path.** A border is an administrative fact, not a travel burden: Kishanganj is 83km from Siliguri in West Bengal and 284km from Patna in its own state.

### Golden table

Each row hand-checked against a map; the tests assert **orderings**, not kilometres, since locality-level geocodes cannot promise an exact distance.

| from | nearer centre | vs | crosses a border? |
|---|---|---|---|
| Darbhanga | **Muzaffarpur** ~52km | Patna ~96km | no |
| Purnia | **Siliguri** ~142km | Patna ~232km | **yes** → West Bengal |
| Kishanganj | **Siliguri** ~83km | Patna ~284km | **yes** → West Bengal |
| Buxar | **Varanasi** ~100km | Patna ~107km | **yes** → Uttar Pradesh |

The last three were **unreachable** under the state model — the same-state rung fired first and never widened.

### The department bug the #99 review found

`hospitals.json` spells departments two ways, because it came from two research passes: snake_case on 12 records, Title Case prose on the other 71. The filter compared raw strings, so **AIIMS Patna's `Radiotherapy` never matched a `radiation_oncology` requirement**. Worse: `gynaecology` and `haematology` — the names `CANCER_TYPE_DEPARTMENTS` used — appear in **no record at all** in that spelling, so those filters matched nothing and were silently skipped by the graceful-degradation branch. Both sides now normalise through one synonym table, and an unrecognised name **fails closed** (matches nothing) rather than matching everything.

## 3. Presentation

New `"distance"` stage renders **`Nearest cancer centres to <city>`**, and each line carries `~N km away (straight-line)` rounded to 10 km, with an explicit prompt instruction never to convert it into a travel time or road distance. **No travel time is claimed anywhere.**

#99's guarantees survive. The widening chain still runs for a city the geocoded table does not know, and distance ordering engages only when **both ends** are geocoded — an origin alone is not a distance signal, so a pool with no coordinates falls back rather than heading a list "Nearest centres" with nothing measured.

## What this deliberately does not do

- **No IP geolocation.** `ll` is not added to `Session`, and nothing in `sessions/` is touched. Per the #103 assessment, `geoip-lite` resolves an Indian Jio range to **Mumbai** — ~1,500km from Darbhanga — so for exactly the rural North Bihar patients this is about it is confidently wrong in a way the city/state gate cannot catch. The text-derived city stays the only origin.
- No `district` capture, no structured catchments, no `nearest_railway` as a distance proxy, no eval cases (so no `cases:update`, and the pinned manifest counters are untouched).

## Tests

- **New `hospital-distance.spec.ts` — 38 cases**, run against the shipped directory rather than a fixture, because the property under test is a property of the real data. Golden orderings; the cross-border assertion; capability-before-distance (a radiotherapy need near Bhagalpur must not return surgery-only Healing Touch — which is both nearest *and* the only centre in its city — with a control proving it *does* lead when no need is stated); the AIIMS Patna regression; fails-closed on unknown departments; ungeocoded-sorts-last; distance tiebreak; fallback for an unknown city; and data integrity — 83 hospitals, every one geocoded or listed unresolved, every coordinate inside India, provenance recorded, every `INDIAN_CITIES` entry geocoded inside India, six Bihar cities within 20km of their hand-checked positions.
- `hospital-directory.service.spec.ts` updated where #99's assertions were **deliberately inverted** by this change — most visibly `orders the Darbhanga candidate set by directory score, not by proximity`, which is now the exact opposite. Purnia is removed from the "HBCH ranked first" list: under distance its nearest centre is Healing Touch Bhagalpur at ~77km. Darbhanga → HBCH, the #95 outcome, is still asserted.
- **Full API suite: 60 suites, 1132 passed, 1 skipped, 0 failed.**
- `npm run build` (`nest build`) clean; `tsc --noEmit` clean.


gautamgauri:
Review status: **HOLD**. Three issues remain on the current head: the planner does not pass `requiredDepartments` into the directory lookup, so the new hard capability filter is not actually guaranteed from the main execution path; the change also adds new LLM instructions under `chat/`, which repository policy requires to be split for SCCF review; and the distance formatter maps values below 5 km (including 0 km locality coordinates) to `~10 km away`, which is numerically misleading. The department normaliser is a good fix. Please wire the capability requirement end-to-end, separate/remove the prompt change, and render very short/low-confidence distances as `within 10 km` or without a number.

gautamgauri:
Codex review addressed — all three findings, in three commits.

**P1-a (capability filter unreachable from the patient flow)** — `0efa64b`. The planner now extracts the treatment a patient names (radiotherapy/sikai, chemo/kimo, surgery/operation; English, Hinglish, Devanagari), maps it to the directory's normalised department names, and passes it as `requiredDepartments`. Three related fixes fell out of it:
- The `hospital_search` signal gate used `\b` around its Devanagari alternatives. JS word boundaries are ASCII-only, so `\bअस्पताल\b` never fired and a Hindi-script hospital question never reached the directory at all — which would have left the Devanagari half of the new extractor unreachable in exactly the way this finding describes. Split out of the boundary group. Only that signal is repaired; the other `\b`-guarded Devanagari alternatives in `detectSignals` share the defect and are untouched (same family as #30) — flagging it rather than widening this PR.
- The stated need and the cancer-type departments are now separate requirement groups, ANDed. Flattened into one set and matched with `some`, "Bhagalpur mein oral cancer ki radiotherapy" was satisfied by `surgical_oncology` and Healing Touch came back for a radiation question.
- Never-empty guarantee kept: if the requirement empties the whole regional pool, the unfiltered pool is returned with `geography.capabilityUnavailable` and `geography.requiredDepartments` set, and the regional heading states that no centre listed offers them.

Tests: planner extraction across all three scripts plus negatives ("chemistry", "cooperation", no need stated); end-to-end through `ExecutionPlannerService` proving the Bhagalpur radiotherapy query — English, Hinglish and Devanagari — never returns Healing Touch, with the Bhagalpur *surgery* query still returning it first as the control.

**P1-b (prompt change needs SCCF review)** — `cbc894f`. The instruction sentence is removed from this PR and proposed on its own in #149 (draft, stacked on this branch, labelled REQUIRES SCCF REVIEW), with the exact text quoted and the reasoning written out. What stays here is data, not guidance: the per-stage headings and the `~N km` line. A regression test pins the boundary — the block must not contain the sentence until that review lands.

**P2 (sub-10km distances inflated to `~10 km`)** — `7f35458`. Under 10 km the line now reads `within 10 km (straight-line)`; 10 km and above rounds to the nearest 10 km as before. The 0 km case you identified — hospital and city sharing a city-confidence coordinate — no longer asserts a distance that was never measured. Tests cover 0 / 0.4 / 3.2 / 9.9 rendering as the bound, 11 / 52.4 / 96 still rounding, and an unmeasured distance still saying nothing.

Full API suite green: 60 suites, 1165 passed, 1 skipped. `npm run build` clean.

gautamgauri:
Re-review on head `7f35458`: **HOLD — the three original findings are fixed, but one new capability-fallback blocker remains.**

The good news: treatment requirements are now extracted and passed end-to-end; the distance-handling prompt sentence was split to SCCF-review draft #149; and sub-10 km values now render truthfully as `within 10 km (straight-line)`. The Purnia generic-search behavior is also defensible: Healing Touch may lead on pure distance when no treatment capability is requested, while radiotherapy/medical-oncology queries correctly filter it out when capable centres exist.

**New blocker:** when the hard capability filter empties the *entire regional pool*, `searchHospitalsWithGeography()` sets `capabilityUnavailable = true` but then deliberately keeps the unfiltered regional pool. Those incapable hospitals still flow through `structuredHospitalResults`; `ChatService` only appends a heading such as `— no centre listed here offers radiation oncology`. That contradicts this PR's own public contract that a centre without the required department is “never offered for that need,” and it makes a safety property depend on the LLM respecting a label while being handed the incapable centres as its authoritative hospital list.

Please fail closed here: if no regional centre satisfies the required capability, return **no incapable regional rows** (while still allowing separately capability-filtered national referrals, if any), or place non-capable centres in a structurally separate field that the patient-facing recommendation path cannot treat as treatment options. Add a regression for a requirement that no regional centre can satisfy, asserting that no incapable centre enters the treatment-option list.

**Stack note:** this PR is also based on #99, not `main`; #99's separate SCCF-gated prompt/travel-caveat issue still has to be cleared before this stack is main-ready.

gautamgauri:
## Capability fail-open — fixed structurally (`552f9d6`, `50547ae`)

Took the **second option: a structurally separate field.** Reasoning below.

### What changed

`HospitalSearchOutcome` gains a third field:

```ts
export interface HospitalSearchOutcome {
  /** The treatment-option list. Every row satisfies the required departments. */
  results: HospitalSearchResult[];
  geography: HospitalSearchGeography;
  /** Regional centres WITHHELD from `results` because not one of them offers
   *  what the search required. Empty in the normal case. */
  nonCapableRegional: HospitalSearchResult[];
}
```

When the hard capability filter empties the entire regional pool,
`searchHospitalsWithGeography()` no longer keeps the unfiltered pool in
`results`. Those rows move to `nonCapableRegional`, and the regional half of
`results` is empty. The national referral pool was already filtered by the same
requirement with no fallback branch, so it still answers.

`ExecutionPlannerService` reads `hospitalOutcome.results` and nothing else, so
`structuredHospitalResults` — the list that becomes the LLM's authoritative
hospital data — can no longer contain a centre that cannot deliver what the
patient asked about. The heading `ChatService` used to append is no longer the
mechanism; it can't even render on that path, because `regional` is empty
whenever `capabilityUnavailable` is set. Its doc comment now says so rather
than claiming to be the safeguard.

### Why the separate field rather than a bare empty set

Both options fail closed for the regional pool. The difference is what is left
to answer with, and I think returning nothing is the worse failure here:

1. **The national arm is the useful answer, and it survives either way.** A
   patient in Bhagalpur asking about proton therapy gets the two national
   centres that actually have it. Zero of the 49 regional centres do. That is
   the real clinical routing — "refer to Patna or Mumbai for this" — and it
   comes out of the existing, already-hard-filtered national path.
2. **An empty directory block invites invention.** `buildHospitalContextBlock()`
   returns `""` for an empty result set, so the whole "use these facts, do not
   mention hospitals not listed" section disappears with it. Empty is the right
   answer when nothing anywhere is capable, but it should not be the answer when
   a capable referral centre exists.
3. **The withheld centres are real and near, and someone will eventually want
   to say so.** They are kept rather than dropped so a future SCCF-reviewed
   surface can say something truthful about them. Nothing renders them today.

The field is deliberately not plumbed into `ExecutionPlan`. There is no path
from it to a prompt, and adding one is a separate reviewed change.

### The regression

New `describe` block in `hospital-distance.spec.ts`, run against the **shipped**
directory, not a fixture. `proton_therapy` is the real case the blocker
describes: offered by 0 of 49 regional centres and 2 of 34 national ones.

- **no incapable regional row for a requirement nothing regional satisfies** —
  `capabilityUnavailable` is true, every row in `results` actually offers proton
  therapy, and the regional slice of `results` is `[]`.
- **the capable national referrals still come back** — `results` is non-empty
  and every row is a national centre that offers it. Fail closed, not fail
  silent.
- **withheld centres stay addressable, off the option list** — non-empty, none
  offers it, disjoint from `results` by id.
- **nothing at all when not even a national centre can serve the need** — an
  unrecognised department yields `results: []`.
- **the ordinary case is untouched** — a Darbhanga radiotherapy query withholds
  nothing and still leads with capable regional centres.
- **end to end through `ExecutionPlannerService`** — with the pools narrowed to
  the blocker's shape (surgery-only Healing Touch Bhagalpur regional, AIIMS
  Patna national, both real records), a radiotherapy query's
  `structuredHospitalResults` excludes Healing Touch and includes AIIMS Patna.

Two existing tests pinned the old fail-open behaviour (`expect(outcome.results.length).toBeGreaterThan(0)`
with `includeNational: false`) and now pin the new contract.

### Tests

```
$ cd apps/api && npx jest --maxWorkers=2 --testPathPatterns=hospital
PASS src/modules/chat/hospital-directory.service.spec.ts
PASS src/modules/chat/hospital-distance.spec.ts
PASS src/common/hospital-directory-file.spec.ts
PASS src/modules/chat/chat.service.hospital-context.spec.ts
Test Suites: 4 passed, 4 total
Tests:       144 passed, 144 total

$ npx jest   # full apps/api suite
Test Suites: 60 passed, 60 total
Tests:       1 skipped, 1171 passed, 1172 total
```

### Wording that is NOT in this PR — needs SCCF review

No prompt text or clinical wording changed here; the structural half only. Two
gaps this change opens, both wording, both for a separate SCCF draft:

1. **The "nothing near you can do this" sentence.** When
   `capabilityUnavailable` is true and only national referrals come back, the
   patient sees national centres with no explanation of why nothing local is
   listed. The block needs a line to the effect of *"no centre in the patient's
   region offers <treatment>; the centres listed are referral centres they
   would travel to."* Structurally it is available — `geography.capabilityUnavailable`
   and `geography.requiredDepartments` are both set.
2. **The empty-directory guard.** When nothing anywhere is capable the hospital
   block is absent entirely, and `navigate-mode.ts` carries no instruction
   against naming centres from the model's own knowledge. This is pre-existing
   (any path with no directory results hits it) but failing closed makes it
   reachable for a new reason, so it is worth a sentence.

Also still true on this head, re-verified: the planner passes treatment
requirements end to end (`execution-planner.service.ts:358`), the distance
prompt sentence is absent (0 occurrences, held for SCCF draft #149), and
sub-10 km renders as `within 10 km (straight-line)` (`chat.service.ts:3211`).

Not rebased — the stack on #99 is untouched.

gautamgauri:
## Codex review triage — all three comments verified at `158b00b`

No code changes were needed: every finding is already fixed on this branch at the head the bot
names. Evidence per comment below; nothing was pushed, `origin/feat/103-hospital-geocoding-distance`
is still at `158b00b`.

| # | Sev | Finding | Verdict | Evidence |
|---|-----|---------|---------|----------|
| 1 | P1 | Move the LLM prompt change through SCCF medical review (`chat.service.ts:3240`) | **ALREADY-ADDRESSED** | The instruction was split out of #148 into the labelled review PR **#149** (`sccf-review/hospital-distance-prompt-instruction`, commit `3edf354`). `git log -S "NEVER convert it into a travel time" -- apps/api/src/modules/chat/chat.service.ts` shows it on that branch only; at this head `chat.service.ts:3254-3261` is a comment recording the hold, and `chat.service.hospital-context.spec.ts:203-216` pins the block free of the sentence. |
| 2 | P1 | Pass required treatment departments from the planner (`hospital-directory.service.ts:46`) | **ALREADY-ADDRESSED** | `execution-planner.service.ts:348` extracts the stated treatment (`extractTreatmentNeeds`) and `:358` passes `requiredDepartments` into `searchHospitalsWithGeography`. Patterns cover English/Hinglish/Devanagari (`:279-303`), with Devanagari written without `\b` (ASCII-only, issue #30). End-to-end: `hospital-distance.spec.ts:192-227` asserts "Which hospital in Bhagalpur for radiotherapy?" (and the Hinglish + Devanagari forms) never returns `healing-touch-bhagalpur`, while the surgery question still does. |
| 3 | P2 | Render sub-10-km distances without inflating them (`chat.service.ts:3243`) | **ALREADY-ADDRESSED** | `Math.max(10, …)` is gone. `chat.service.ts:3238-3243` renders `within 10 km (straight-line)` below 10 km and `~N km away (straight-line)` above, so a same-city 0 km never becomes "~10 km". It is the only distance renderer in the codebase (`grep -rn "distance_km" apps/api/src`). Covered by `chat.service.hospital-context.spec.ts` (sub-10 km bound case). |

### Tests

- `cd apps/api && NODE_OPTIONS=--max-old-space-size=2048 npx jest --runInBand` → **74 suites passed, 1525 passed / 1 skipped / 1526 total**, exit 0. The one skip is pre-existing and environmental: `reference-chunk-filter.spec.ts:155` gates a KB-wide sanity block on the local PDQ corpus, which is not present in this checkout.
- Targeted: `npx jest --runInBand --testPathPatterns='(hospital-distance|hospital-context|execution-planner)'` → 3 suites, 117 tests passed.
- `apps/web` untouched by this PR (`git diff --stat origin/main...HEAD` lists no web files), so the vitest suite was not run.

### Notes for the human reviewer (not fixed here, by scope)

- The per-hospital `within 10 km / ~N km away (straight-line)` text and the stage headings
  (`Nearest cancer centres to X`, `— no centre listed here offers …`) are still new strings entering
  the LLM's hospital block. This PR treats them as data labels rather than instructions (see
  `da0c499`); if SCCF wants that line drawn differently, they belong in #149 too. **Your call.**
- `plan()` only performs the hospital lookup when `category === "NAVIGATION"`; a radiotherapy
  question routed to another category still gets no capability filter. Pre-existing, out of scope
  for these comments.

gautamgauri:
Cleared to merge ahead of its wording, with the deploy held instead.

`da0c499` on this branch held the distance instruction back into #149 per AGENTS.md §1.3, so merging this alone renders `~N km away (straight-line)` into the prompt block with nothing telling the model what that figure is. Deploys here are manual `gcloud builds submit`, so merging does not ship — the only real risk was a later deploy riding along.

That risk is now closed mechanically: **Check 9 of the deploy preflight** (`.claude/commands/deploy.md`, commit 53e32eb) fails the deploy when `chat.service.ts` contains `(straight-line)` without `STRAIGHT-LINE distance, already rounded`. It arms itself when this PR lands and disarms when #149 lands — no manual bookkeeping.

**Blocked on for deploy, not for merge:** #149 (distance-handling instruction) and #159 (cross-border travel caveat). Both need SCCF review.

PR #149, opened 2026-09-13
REQUIRES SCCF REVIEW: distance-handling instruction in the hospital prompt block

gautamgauri:
**Draft — blocked on SCCF human/medical review. Do not merge without it.**

Stacked on #148. Contains exactly one prompt change, split out of that PR per AGENTS.md §1.3 ("Do not edit medical content, prompts, or clinical policy on your own authority… Propose them in a separate, clearly-labeled PR"), raised by the Codex review on #148.

## The exact text

One sentence appended to the `=== VERIFIED HOSPITAL DIRECTORY DATA ===` block in `buildHospitalContextBlock()`:

> Where a distance is given it is a STRAIGHT-LINE distance, already rounded. Repeat it as written if you mention it. NEVER convert it into a travel time, a road distance, or a journey duration — the road route is longer and the time depends on connections this data does not contain.

Nothing else changes. The diff is that sentence plus the unit test that pins it.

## Why it exists

#148 geocodes the directory and puts a distance on each hospital line (`~50 km away (straight-line)`, or `within 10 km (straight-line)` under ten). The figure itself is a data label — the kind of factual heading that stays in #148 — but an unguarded number in a prompt invites the generator to do two things the data does not support:

1. **Call a far centre "nearby."** The block can legitimately list a centre 90–150 km away as the nearest capable one. Without the instruction there is nothing stopping the model framing that as close by to a patient who is deciding where to travel for treatment.
2. **Turn kilometres into hours.** These are great-circle distances over locality-level geocodes. The road route is always longer, and journey time depends on rail and road links the directory holds nothing about — Darbhanga to Muzaffarpur is well connected, while other 100 km hops in the same region are four hours. A confidently invented "about an hour from you" is the failure mode this sentence is written against.

The existing `travelCaveat` (from #99, already reviewed) covers only the adjacent-state and unfiltered rungs. The distance-ordered rung — the new one — has no such guard.

## What SCCF is being asked to decide

- Is this the right instruction, in the right register, for a patient-facing navigation answer?
- Should it go further and forbid mentioning the distance at all in the reply, rather than permitting it to be repeated as written?

## Tests

`apps/api/src/modules/chat/chat.service.hospital-context.spec.ts` — the assertion that #148 uses to keep the sentence *out* is flipped here to assert it is present. Full API suite green (1165 passed, 1 skipped, 60 suites); `npm run build` clean.


gautamgauri:
Review status: **KEEP DRAFT / READY FOR SCCF HUMAN REVIEW**. This is correctly split from #148 and is narrowly scoped to one model instruction about how to handle the structured straight-line distance field. Technically, the instruction is consistent with the data contract: repeat the rounded straight-line value if used, and do not invent road distance or travel time. I do not see an engineering blocker, but under AGENTS.md §1.3 this should remain draft until SCCF records the human/medical approval. It should also follow #148 in the stack.

gautamgauri:
This PR is now a **deploy blocker**, not just a review item.

#148 (the distance-ranking mechanism) is being merged to `main` ahead of this wording. Check 9 of the deploy preflight (`.claude/commands/deploy.md`, commit 53e32eb) will refuse to deploy `suchi-api` until this lands — the mechanism renders a kilometre figure to the generator, and this PR carries the language that makes it safe to say out loud.

Nothing ships to patients until SCCF signs off here. Routing per the usual path (Divya → Dr. Amit).

gautamgauri:
Superseded by #181 — same branch, rebased onto `main`, same single commit and same review question.

This PR was auto-closed when #148 merged and `feat/103-hospital-geocoding-distance` was deleted. GitHub will not retarget a closed PR's base, and will not reopen one whose base branch is gone, so the work moved to a new number rather than being reconstructed.

PR #156, opened 2026-09-14
fix(chat): route by who is speaking before what is wrong (#154)

### LINKED ISSUE #154: Patient-state routing: any caregiver query mentioning a stage or "diagnosed with" is classified POST_DIAGNOSIS, so caregivers get the patient contract
Split out of #152, whose "Observation, **unverified**" note this confirms and widens. The display-boundary strip is #153; this is the routing defect underneath it.

## Mechanism (verified)

`PatientStateService.detect()` (`apps/api/src/modules/chat/patient-state.service.ts:143`) is first-match-wins over a fixed tier order:

```
URGENT > POST_DIAGNOSIS > CAREGIVER > SIDE_EFFECTS > SYMPTOMATIC > INFORMATIONAL
```

Each tier `return`s as soon as any of its patterns match, so **CAREGIVER is never evaluated when a POST_DIAGNOSIS pattern also matches**.

The overlap is not incidental. The POST_DIAGNOSIS patterns are *disease-fact* markers — `\bstage\s+[1-4IV]+\b`, `\bdiagnosed\s+with\b`, `\bbiopsy\s+report\b`, `\bgrade\s+[1-3]\b` — all equally true whether the speaker is the patient or a relative. The CAREGIVER patterns (`\bmy\s+(father|mother|…|uncle|…)\b`) are the only ones carrying *who is speaking*. Checking the "what" tier before the "who" tier is backwards for exactly the queries where both are present.

## Reproduction

Instantiating the service directly and calling `detect()`:

| query | detected | correct |
|---|---|---|
| `My uncle has stage 4 lung cancer. Exactly how many months does he have left?` | `post_diagnosis` (medium) | `caregiver` |
| `My father was diagnosed with breast cancer, what should I do?` | `post_diagnosis` (medium) | `caregiver` |
| `My uncle has lung cancer. How many months does he have left?` | `caregiver` (high) | ✓ |
| `I have stage 4 lung cancer, what are my options?` | `post_diagnosis` (medium) | ✓ |

Row 1 is the #152 QA query. Row 2 matters more: **"my father was diagnosed with…" is the single most natural way a caregiver opens**, and it misroutes too. The only caregiver queries that classify correctly today are those that mention *no* clinical fact.

Note the confidence signal is inverted as well — the misroutes are `medium` off one pattern, while the correct caregiver detections are `high`.

## Why it matters

The contracts differ substantially (`llm.service.ts:597-608` vs `:626-652`). A caregiver routed to POST_DIAGNOSIS is:

- **addressed as the patient** — the contract opens `"I understand receiving a [cancer type] diagnosis can be overwhelming"`, said to someone whose uncle is ill
- **denied the caregiver sections** — the ≥5 caregiver action steps, the hospital preparation checklist, and the support resources (Indian Cancer Society `1800-22-1951`, Vandrevala `9999666555`) exist only in the CAREGIVER contract
- given "questions *the patient* should ask" rather than questions to ask on the patient's behalf

So the people most likely to be organising care get the least navigational help, and the helpline numbers never reach them.

## Fix shape (not yet decided)

Score the tiers instead of returning on first match, or check CAREGIVER's relation+cancer-context test *before* POST_DIAGNOSIS and let it win when both fire. The second is smaller, but either way a first-person marker (`I have`, `my report`, `मुझे`) must still beat a relation word so `"my father had it, and now I have stage 2"` stays POST_DIAGNOSIS.

Needs a decision on that precedence rule and a fixture set covering both directions before coding.

## Related
- #152 — the delivered scaffolding this sits under
- #153 — the display-boundary strip (does not address this)

gautamgauri:
Closes #154. Deterministic precedence fix, not a redesign into scoring.

## The bug

`detect()` returns on the first matching tier, and POST_DIAGNOSIS was checked **before** CAREGIVER. Its markers — `stage [1-4]`, `diagnosed with`, `biopsy report`, `my report` — are **disease facts**, equally true whether the speaker is the patient or a relative. The CAREGIVER patterns are the only ones carrying *who is asking*.

| query | before | after |
|---|---|---|
| `My uncle has stage 4 lung cancer…` | `post_diagnosis` | **`caregiver`** |
| `My father was diagnosed with breast cancer, what should I do?` | `post_diagnosis` | **`caregiver`** |
| `My father had cancer and now I have stage 2` | `post_diagnosis` | `post_diagnosis` ✓ |
| `I have stage 4 lung cancer, what are my options?` | `post_diagnosis` | `post_diagnosis` ✓ |

The second row is the most natural way a caregiver opens. In practice the only caregiver queries that classified correctly were those mentioning no clinical fact at all.

## Why it isn't cosmetic

The contracts differ materially (`llm.service.ts:597-608` vs `:626-652`). A caregiver on the patient contract is **addressed as though they are the ill one** — the contract opens `"I understand receiving a [cancer type] diagnosis can be overwhelming"` — and loses the caregiver-only material: the action steps, the hospital preparation checklist, and the support helplines (Indian Cancer Society `1800-22-1951`, Vandrevala `9999666555`). The people most likely to be organising care were getting the least navigational help, and the helpline numbers never reached them.

## The fix

Precedence is now:

```
URGENT  >  first-person disease ownership  >  CAREGIVER  >  POST_DIAGNOSIS  >  rest
```

A new `FIRST_PERSON_DISEASE_OWNERSHIP` set is the tie-breaker in the other direction, so an explicit claim on the illness still wins over a relation word. Nothing else about the tiers changed.

## Two script gaps the fixtures surfaced

Both are fixed here, because otherwise the Hinglish and Devanagari fixtures you asked for could not pass:

1. **Hinglish** — the relation rule required the *English* `my` (`/\bmy\s+(bhai|maa|papa…)/`), so `mere papa`, `meri maa`, `mera bhai` matched nothing and fell through to `INFORMATIONAL`.
2. **Devanagari** — there were **no** relation or cancer-context patterns at all, so `मेरे पिता को कैंसर है` was never a caregiver.

Both are written without `\b` throughout: JavaScript word boundaries are ASCII-only and meaningless against Devanagari, a bug this project has shipped before. There's an explicit test asserting these don't regress to `INFORMATIONAL`.

The relation nouns and ownership nouns are deliberately disjoint, so `meri maa` reads as a relation while `meri report` still reads as the speaker's own.

## Tests

**25 fixtures** across English, Hinglish and Devanagari. Beyond the routing table above, they pin the things that must *not* move: URGENT still outranks everything, a relation word with no cancer context does not become CAREGIVER, `What is a biopsy?` stays informational, and first-person symptom reports stay SYMPTOMATIC.

Suite **1377 passing** (was 1352).

## Note

This changes which contract a caregiver query receives. It does **not** change any contract wording — that is clinical prompt surface and goes through SCCF review separately.


gautamgauri:
Both review findings are fixed in code, not argued away. Commits `a0079b4` (fix) and `3cac56c` (fixtures).

The two findings pull in opposite directions — P1 wants the first-person set narrowed, P2 wants it to stop being the sole gate — so narrowing the one list would have made P2 worse. The root cause is that one list was answering two different questions. It is now two lists with deliberately opposite breadth:

- **`FIRST_PERSON_SELF_REFERENCE`** (`patient-state.service.ts:176`) — **WHO is speaking.** Broad on purpose. It gates CAREGIVER only. A false positive costs no more than the routing this file had before #154.
- **`FIRST_PERSON_DIAGNOSIS_OWNERSHIP`** (`patient-state.service.ts:190`) — **WHAT is confirmed.** Narrow on purpose. It is the only first-person route into POST_DIAGNOSIS, whose contract acknowledges a diagnosis and enumerates treatments.

`detect()` now gates CAREGIVER on `speaksForSelf` (`:305`) and POST_DIAGNOSIS on `ownsDiagnosis` (`:321`) — `patient-state.service.ts:290-292`.

I did not add a new `PatientState`. An awaiting-diagnosis state would need its own response contract, and prompt/clinical-surface wording goes through the separate SCCF-review PR by repo policy. See "what I did not do" below.

---

### Finding 1 (P1) — a pending procedure is not a diagnosis

Removed from the diagnosis set: bare `my biopsy`, `my oncologist`, `my surgery`, `my results`, `my report`, and `I have a (lump|mass|tumour)` — every one of those is equally true of somebody still under workup. Also removed the Hinglish `meri biopsy` and the Devanagari `मेरी बायोप्सी` / `मेरी सर्जरी` / `मुझे गाँठ` alternatives, since the sets are per-script and narrowing only English would just move the bug.

What still counts as a diagnosis: stated disease (`I have cancer/stage N/lymphoma…`), `I was diagnosed`, `my cancer is/was`, care that only follows a diagnosis (`my diagnosis|chemo|radiation|mastectomy`), receptor status, `I'm a patient/survivor`, and the script equivalents (`mujhe cancer`, `मुझे कैंसर`, `मेरे इलाज`). "My biopsy report says grade 2" still routes POST_DIAGNOSIS through the pre-existing result-bearing patterns at `:63-76`.

`My biopsy is scheduled tomorrow; what should I expect?` now reaches INFORMATIONAL — no contract, so nothing asserts a diagnosis.

Covered by `patient-state.speaker-ownership.spec.ts` → **"a pending procedure is not a diagnosis (PR #156 review, P1)"**:
- `is never told they are diagnosed: …` (8 cases, incl. the finding's exact string, "I am waiting for my biopsy results.", an oncology appointment, a scheduled operation, `Meri biopsy kal hai…`, `मेरी बायोप्सी कल है…`, `मेरी सर्जरी अगले हफ्ते है`) — asserts **never POST_DIAGNOSIS**, i.e. on the harm rather than on any one replacement state
- `the flagship case lands on a contract that assumes nothing` — pins the exact string to INFORMATIONAL so a future change of landing state is a visible decision
- `a lump one is still having looked at is a symptom, not a diagnosis` — `I have a lump in my breast` → SYMPTOMATIC
- `narrowing did not break actual diagnoses` — 4 genuine post-diagnosis strings still POST_DIAGNOSIS

### Finding 2 (P2) — patient routing survives unlisted first-person forms

The CAREGIVER early return is no longer gated on proof of disease. `My wife wants to understand my treatment plan` now routes POST_DIAGNOSIS: `my treatment` is self-reference, which suppresses CAREGIVER without itself claiming a diagnosis, and the pre-existing `treatment plan` pattern then does the routing. The discriminator is the possessive, not the relation word — `My wife wants to understand **her** treatment plan` is still CAREGIVER, asserted.

Covered by **"a patient who mentions a relative stays a patient (PR #156 review, P2)"**:
- `routes to POST_DIAGNOSIS, not CAREGIVER: …` (6 cases: the finding's exact string, `my chemo schedule`, `my oncologist recommended`, `my radiation sessions`, `Meri patni mera treatment plan samajhna chahti hai`, `मेरी पत्नी मेरे इलाज के बारे में जानना चाहती है`)
- `the relative's own illness still reads as CAREGIVER`
- `self-reference alone does not manufacture a diagnosis` — "My mother had breast cancer. I found a lump last week." → SYMPTOMATIC, the case that proves the two sets are genuinely separate

### Verification

Both findings reproduce. Running the new fixtures against the pre-fix classifier (`f2ad739`) fails **13** of them — 9 from finding 1, 4 from finding 2 — and all 44 pass on HEAD.

- `npx jest --testPathPatterns=patient-state` → **44 passed, 44 total** (25 pre-existing #154 fixtures unchanged and still green)
- full `apps/api` suite → **Test Suites: 72 passed, 72 total; Tests: 1 skipped, 1395 passed, 1396 total**
- `tsc --noEmit` clean

### What I did not do

- **No new `PatientState`, and no contract wording touched.** "Awaiting diagnosis" is a genuinely distinct state and the enum has no member for it, but adding one means writing a new response contract in `llm.service.ts` — clinical surface, which repo policy routes through a separate SCCF-review draft PR. So these queries fall through to the existing assumption-free tiers instead. That is safe (INFORMATIONAL has no contract; SYMPTOMATIC's contract explicitly forbids assuming a diagnosis) but it is not *helpful* — a person whose biopsy is tomorrow gets general information rather than "here is what the procedure is like, here is when results come, here is what to ask". **Recommended follow-up: an AWAITING_DIAGNOSIS state plus its contract, drafted for SCCF review.** I'd suggest it should cover what the procedure involves, the result timeline, questions to ask, and explicitly that most biopsies do not find cancer — and must not name treatments.
- **Devanagari symptom reports still under-route.** With `मुझे गाँठ`/`मुझे ट्यूमर` correctly out of the diagnosis set, "मुझे गाँठ है" lands INFORMATIONAL rather than SYMPTOMATIC, because `SYMPTOMATIC_FIRST_PERSON` and `SYMPTOM_KEYWORDS` (`:196-217`) are ASCII-only — there are no Devanagari symptom patterns at all. Strictly safer than the POST_DIAGNOSIS this PR briefly gave them, but it is a real gap and it predates this PR. Worth its own issue.
- `my report` at `:64` and `what does my report/biopsy` at `:76` are long-standing POST_DIAGNOSIS patterns, untouched here. `meri report` / `मेरी रिपोर्ट` are kept in the diagnosis set for parity with them; changing that class of pattern is a separate call.

gautamgauri:
Re-review on current head `3cac56c`: the two earlier routing findings are addressed in the diff. The implementation now separates broad first-person self-reference from narrow diagnosis ownership, so a pending biopsy/appointment does not itself unlock the POST_DIAGNOSIS contract, while forms such as “my treatment plan” preserve patient ownership when a relative is mentioned. The new English/Hinglish/Devanagari fixtures cover both directions, and API build verification is green. I do not see a remaining correctness or multilingual blocker on this head; this looks ready to merge.

